PR #35, opened 2015-11-18
Added readme notice about installing Apollo from source

jpettersson:
Apollo is not actually distributed yet. I added a note about installing from source.

@rouzwawi


rouzwawi:
:+1:

PR #37, opened 2015-11-19
some tweaks to the front page

pettermahlen:
- Moving around the order of api/core/http-service because I think it's clearer that way
- Changing some descriptions a bit, also for clarity


rouzwawi:
Conflicted with the other spelling fixes. Please rebase.

PR #40, opened 2015-11-22
Use APOLLO as default env var prefix

### LINKED ISSUE #39: apollo-core looking for environment variables in the SPOTIFY_X_Y namespace.
While it works perfectly for spotify internal services mandating that for other peoples services feels a bit wrong and is likely a source for confusion (ehh, why is this called spotify, huh?). Perhaps default to APOLLO_X_Y or $SERVICENAME_X_Y?

Referenced here:
https://github.com/spotify/apollo/blob/master/apollo-core/README.md

Implemented here:
https://github.com/spotify/apollo/blob/master/apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java#L421

rouzwawi:
fixes #39

### DIFF
--- a/apollo-environment/src/test/java/com/spotify/apollo/environment/ApolloEnvironmentModuleTest.java
+++ b/apollo-environment/src/test/java/com/spotify/apollo/environment/ApolloEnvironmentModuleTest.java
@@ -136,7 +136,7 @@ public class ApolloEnvironmentModuleTest {
   public void shouldSetUpConfig() throws Exception {
     final AtomicReference<Environment> envReference = new AtomicReference<>();
 
-    final Map<String, String> env = ImmutableMap.of("SPOTIFY_APOLLO_BACKEND", "my-domain");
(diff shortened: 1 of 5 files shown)


apollo-environment/src/test/java/com/spotify/apollo/environment/ApolloEnvironmentModuleTest.java
```diff
@@ -136,7 +136,7 @@ private static void validateEnv(Environment e) {
   public void shouldSetUpConfig() throws Exception {
     final AtomicReference<Environment> envReference = new AtomicReference<>();
 
-    final Map<String, String> env = ImmutableMap.of("SPOTIFY_APOLLO_BACKEND", "my-domain");
+    final Map<String, String> env = ImmutableMap.of("APOLLO_APOLLO_BACKEND", "my-domain");
```
rouzwawi: Arguably a bit unfortunate

PR #42, opened 2015-11-23
Set version to 1.0.0-SNAPSHOT

zalenski:
@rouzwawi 

Any reason apollo-example and the example apps can't be submodules?
Now they won't get built and tested in a normal mvn verify.

### DIFF
--- a/apollo-example/pom.xml
+++ b/apollo-example/pom.xml
@@ -9,8 +9,8 @@
         those dependencies transitively bring in.
     </description>
     <groupId>com.spotify</groupId>
-    <artifactId>apollo-example</artifactId>
-    <version>1.0.0-SNAPSHOT</version>
+    <artifactId>apollo-example-app</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <dependencyManagement>
--- a/apollo-example/pom.xml
+++ b/apollo-example/pom.xml
@@ -10,7 +10,7 @@
     </description>
     <groupId>com.spotify</groupId>
     <artifactId>apollo-example</artifactId>
-    <version>1.0.0-rc3-foss-SNAPSHOT</version>
+    <version>1.0.0-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <dependencyManagement>
@@ -18,7 +18,7 @@
             <dependency>
                 <groupId>com.spotify</groupId>
                 <artifactId>apollo-bom</artifactId>
-                <version>1.0.0-rc3-foss-SNAPSHOT</version>
+                <version>1.0.0-SNAPSHOT</version>
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
--- a/examples/calculator/pom.xml
+++ b/examples/calculator/pom.xml
@@ -6,7 +6,7 @@
     <description>A simple calculator application</description>
     <groupId>com.spotify</groupId>
     <artifactId>calculator-service</artifactId>
-    <version>0.0.1-SNAPSHOT</version>
+    <version>1.0.0-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <dependencyManagement>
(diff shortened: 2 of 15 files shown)


rouzwawi:
I think it had something to do with the way we imported the bom globally. But with the current setup of importing it everywhere, it might be possible to include them in the main project.

examples/calculator/pom.xml
```diff
@@ -6,7 +6,7 @@
     <description>A simple calculator application</description>
     <groupId>com.spotify</groupId>
     <artifactId>calculator-service</artifactId>
-    <version>0.0.1-SNAPSHOT</version>
+    <version>1.0.0-SNAPSHOT</version>
```
rouzwawi: This one should probably not be versioned with the main artifacts. The example is not supposed to be published, so it should stay on 0.0.1-SNAPSHOT

zalenski: I've always found this just slows you down. You would like the examples to be built and tested, so they should be built with the main project. And you'll have to update the examples so they compile with new versions, so they might as well follow the version they're an example of…? If they are built with the main project, you get everything updated for you with versions:set and release:prepare. Keeping them on a separate version is more cumbersome.

(But they should be tied to the apollo-http-service version they're using when we'll start versioning the api separate from modules and assemblies.)

zalenski: Btw, how do you prevent some modules from being published, if you still want them verified in your builds?

Using `maven-deploy-plugin` `<skip>true</skip>` is suggested in [maven faq#skip](https://maven.apache.org/plugins/maven-deploy-plugin/faq.html#skip)

rouzwawi: But this is the version of the module itself, not the apollo dependencies it has.

examples/calculator/pom.xml
```diff
@@ -6,7 +6,7 @@
...
     <packaging>jar</packaging>
 
     <dependencyManagement>
```
rouzwawi: The bom version a few lines below should change instead.

examples/calculator/pom.xml
zalenski: Yes, but the cumbersomeness have come from having to exclude it in every mvn release:prepare, and it refusing to keep the snapshot version when asking for what version it should set that artifact to.

I guess we should start separating more of the versioning. But does that mean we should have multiple builds set up, for each part of the project? (api, impl, http module, assembly, examples)

apollo-example/pom.xml
```diff
@@ -10,7 +10,7 @@
     </description>
     <groupId>com.spotify</groupId>
     <artifactId>apollo-example</artifactId>
-    <version>1.0.0-rc3-foss-SNAPSHOT</version>
+    <version>1.0.0-SNAPSHOT</version>
     <packaging>jar</packaging>
 
     <dependencyManagement>
```
rouzwawi: Same here, a few lines below.

examples/calculator/pom.xml
zalenski: fixed

apollo-example/pom.xml
zalenski: fixed

examples/calculator/pom.xml
zalenski: I think all the examples are 1.0.0-SNAPSHOT, all separately :p

rouzwawi: Fair enough :) as long as they have artifact ids that make it clear that they are not anything interesting to depend on.

rouzwawi:
:+1:

rouzwawi:
Do you want to add them as modules in the main pom as well?

examples/calculator/pom.xml
zalenski: But currently they're separate, and not built, so there'll be no problems with release:prepare either. I'll go with 0.0.1-SNAPSHOT (for both apollo-example (which we may remove/move to an example?) and example/calculator).

zalenski:
> Do you want to add them as modules in the main pom as well?

Or do we want to separate them?

> > I guess we should start separating more of the versioning. But does that mean we should have multiple builds set up, for each part of the project? (api, impl, http module, assembly, examples)

rouzwawi:
This looks good for now. We can decide to either include or separate later.

PR #45, opened 2015-11-24
No example deploys, Include examples as sub modules

zalenski:
maven-deploy-plugin skip deploys,
build examples as submodules

this brings back some cumbersomeness when using versions:set or release:prepare, if example versions do not follow the main project version, hence the

```
      <!-- TODO: should separate ci builds be set up for each independent group of artifacts? -->
      <!-- (api, impl, modules, assemblies, examples, ...) -->
```

### DIFF
--- a/apollo-example/pom.xml
+++ b/apollo-example/pom.xml
@@ -81,6 +81,18 @@
                     </compilerArgs>
                 </configuration>
             </plugin>
+
+            <!-- Do not deploy, this is just an example -->
+            <!-- TODO: Remove or move to one example -->
+            <!-- TODO: set up ci to build and verify this -->
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-deploy-plugin</artifactId>
+              <version>2.8.2</version>
+              <configuration>
+                <skip>true</skip>
+              </configuration>
+            </plugin>
         </plugins>
     </build>
 
--- a/pom.xml
+++ b/pom.xml
@@ -25,6 +25,13 @@
         <module>modules/slack</module>
         <module>modules/jetty-http-server</module>
         <module>modules/okhttp-client</module>
+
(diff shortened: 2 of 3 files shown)


apollo-example/pom.xml
```diff
@@ -81,6 +81,18 @@
...
                 </configuration>
             </plugin>
+
+            <!-- Do not deploy, this is just an example -->
+            <!-- TODO: Remove or move to one example -->
+            <!-- TODO: set up ci to build and verify this -->
+            <plugin>
+              <groupId>org.apache.maven.plugins</groupId>
+              <artifactId>maven-deploy-plugin</artifactId>
+              <version>2.8.2</version>
```
pettermahlen: You shouldn't have to specify the version of the deploy plugin here - it should be defined in some parent pom instead, though maybe it isn't.

pettermahlen:
I think it's OK to build the examples with the main build, as long as the don't get deployed - or installed, I guess. So maybe the install plugin should be disabled as well.

pom.xml
```diff
@@ -25,6 +25,13 @@
         <module>modules/slack</module>
         <module>modules/jetty-http-server</module>
         <module>modules/okhttp-client</module>
+
+        <!-- include examples as modules, to get them built and verified -->
+        <!-- NOTE: they configure the maven-deploy-plugin to skip deployment -->
+        <!-- TODO: should separate ci builds be set up for each independent group of artifacts? -->
+        <!-- (api, impl, modules, assemblies, examples, ...) -->
+        <module>apollo-example</module>
+        <module>examples/calculator</module>
```
rouzwawi: These should have their counterpart `<parent>` declaration in the sub modules.

rouzwawi:
@zalenski is this still relevant? I've not seen any problems related to this with the current setup.

zalenski:
> is this still relevant?

no, probably not (…one year later). closing this

PR #48, opened 2015-11-24
Enable japicmp plugin on select modules

rouzwawi:
I propose that we enable this when releasing 1.0.0.

The plugin is only added to some of the modules. Those with a more directly publicly available packages.


rouzwawi:
This is good to merge now

dflemstr:
:+1:

PR #49, opened 2015-11-25
Rename/Move apollo-example to example/brewery

aes:
@rouz @zalenski Trivial rename, but I wanted to be sure you're not using it the old names for tests anywhere.


rouzwawi:
:+1:

PR #52, opened 2015-11-29
Add map and flatMap to (A)SyncHandler

rouzwawi:
While experimenting with writing some middlewares for serialisation/deserialisation, I realised that the handler types are [Reader Monads][1].

Adding these methods helps making middleware implementations much more readable:

```java
<P> Middleware<PayloadResponseHandler<P>, SyncHandler<Response<ByteString>>> forType(Class<P> payloadClass) {
  return innerHandler ->
      deserialize(payloadClass)
          .flatMap(resp -> ctx ->  innerHandler.apply(ctx).apply(resp.payload().get()))
          .map(this::serialize);
}

<P, T> Middleware<PayloadArbitraryHandler<P, T>, SyncHandler<Response<ByteString>>> forType(Class<P> payloadClass) {
  return innerHandler ->
      deserialize(payloadClass)
          .flatMap(resp -> ctx ->  innerHandler.apply(ctx).apply(resp.payload().get()))
          .map(this::ensureResponse)
          .map(this::serialize);
}
```

The signature of `deserialize` is
```java
<P> SyncHandler<Response<P>> deserialize(Class<P> payloadClass)
```

`map` and `flatMap` at least makes it easier to add functions to a request handler chain by not having to explicitly pass along the `RequestContext` in places where it is not needed. But there's a piece missing in being able to map the payload of a response, if present. I will open a separate PR to address that since it's orthogonal to the handler api. see #53 

[1]: http://adit.io/posts/2013-06-10-three-useful-monads.html#the-reader-monad

### DIFF
--- a/apollo-api-impl/src/main/java/com/spotify/apollo/route/Middlewares.java
+++ b/apollo-api-impl/src/main/java/com/spotify/apollo/route/Middlewares.java
@@ -27,7 +27,6 @@ import com.spotify.apollo.StatusType;
 import com.spotify.apollo.serialization.AutoSerializer;
 
 import java.util.Optional;
-import java.util.function.Function;
 
 import okio.ByteString;
 
@@ -61,9 +60,8 @@ public final class Middlewares {
   public static AsyncHandler<Response<ByteString>> httpPayloadSemantics(
       AsyncHandler<Response<ByteString>> inner) {
 
-    return requestContext ->
-        inner.invoke(requestContext)
-            .thenApply(applyHttpPayloadSemantics(requestContext.request()));
+    return inner.flatMapSync(resp -> ctx -> applyHttpPayloadSemantics(ctx.request(), resp));
+
   }
 
   /**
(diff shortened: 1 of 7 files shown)


pettermahlen:
TBH, I'm not convinced it improves readability. At least, I don't understand the code above or what reader monads are (despite spending 10-15 minutes trying to understand the linked article). I understand it is shorter, and I'm sure if I practice functional programming a bit more, I will understand. Right now, I don't, and it's my guess that many/most Java programmers also don't.

That doesn't mean opening up for people to use map/flatMap is a bad idea. It's not the same as forcing people to do it. But I think that it's not going to improve readability for the general public. It can reduce the amount of code.

Some questions we should consider given that we're already receiving feedback about how Middlewares are hard to understand, and this makes them still harder, I think:

- is this a case of 'when in doubt, leave it out'? 
- is there a risk of 'functional snobbery' that goes counter to 'make this a pragmatic, programmer-friendly framework'?

An alternative way to look at this is, I think, something like "many Java programmers are not yet used to thinking in these terms, but they will learn. Middlewares and functional idioms like these will soon be obvious, so it's best to put them in there already." I'm not sure.

pettermahlen:
A completely separate note: I think it's a good idea to move TestContext into apollo-test and make that a generally available class, with some additional 'sugar' to be able to include a ```(Stub)Client```.

mbruggmann:
We had a promising discussion about alternative approaches offline, so I'd vote for holding off on this one for now.

rouzwawi:
I kind of agree on that it adds an additional layer of abstraction on the implementation which makes it less understandable. The thing @mbruggmann and I discussed was how to approach my use case in a different way (not using middleware) to make it more readable.

That said, I still believe that these composition helpers are useful on `(A)SynchHandler` when writing middleware. So I would still like to merge this PR.

pettermahlen:
I think while this does add some conceptual weight to the API, it also adds power. So :+1: from me!

mbruggmann:
Sure, let's add this as well! :+1:

Issue #11, opened 2015-11-12
Add map of headers to Request


cjhawley:
I think it would be convenient if Request accepted a `Map<String, String> headers` so code can go from 

```   
Request request = Request.forUri(uri, "GET");
request = request.withPayload(ByteString.encodeUtf8(json));
for (Map.Entry<String, String> h : context.getHeaders().entrySet()) {
  request = request.withHeader(h.getKey(), h.getValue());
}
```

to

```
Request request = Request.forUri(uri, "GET");
request = request.withPayload(ByteString.encodeUtf8(json));
request = request.withHeaders(context.getHeaders());
```

An alternative, would be to allow a RequestBuilder that lets me avoid the reassignment as well.

PR #55, opened 2015-11-30
Add withHeaders(Map) to Response

rouzwawi:
fix for #11

### DIFF
--- a/apollo-api/src/main/java/com/spotify/apollo/Response.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/Response.java
@@ -58,6 +58,14 @@ public interface Response<T> {
   Response<T> withHeader(String header, String value);
 
   /**
+   * Creates a new {@link Response} based on this, but with additional headers.
+   *
+   * @param headers  Map of header name-values to add
+   * @return A response with the added headers
+   */
+  Response<T> withHeaders(Map<String, String> headers);
+
+  /**
    * Creates a new {@link Response} based on this, but with a different payload. Status code,
    * headers, etc., are copied over. To clear out the payload, one can pass in {@code null}.
    *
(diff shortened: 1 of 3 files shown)


rouzwawi:
@cjhawley

pettermahlen:
I wonder if we should have two methods ```withAddedHeaders``` and ```withHeaders```. As it is, there's no way to set the headers for the response to some exact map, and also no way to clear a header.

rouzwawi:
Is it common that you want all existing headers to be cleared out?

I'm more on the line of reasoning that, if I get a response instance as an argument a method/function, then I expect it to have data in it that is there for a reason, so I would only add to it. If I really needed to discard things, I would have to construct a new response from scratch. But I have not run into the need to clear out headers.

pettermahlen:
I agree that would be the majority use case. I wonder if there are enough exceptions to warrant adding another method. I guess it comes down to writing something like this:

```
  Response<T> incoming = getIncomingResponse(); 
 
  Response<T> realResponse = Response.forStatus(incoming.status())
    .withHeaders(removeHeaderIDontWant(incoming.headers());

  if (incoming.payload().isPresent()) {
    return realResponse.withPayload(incoming.payload.get());
  }

  return realResponse;
```

pettermahlen:
On another note, I just noticed #11 is about Request, not Response

rouzwawi:
oh, so this does not resolve #11 :(

On the headers handling, yes that would be something to write if you're implementing a proxy to a different service. In most cases there will probably be some domain object in between so that the two responses would be a bit separated in the code.

rouzwawi:
oops, I accidentally pushed this to master when pushing ca941c1dc6a284d8815ff207828285265eb151ac

PR #56, opened 2015-12-02
Various improvements to Request

pettermahlen:
Fixes a couple of reported issues, improves javadoc comments, and ensures that the maps exposed by Request are safe/not modifiable.

### DIFF
--- a/apollo-api/src/main/java/com/spotify/apollo/RequestValue.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/RequestValue.java
@@ -20,6 +20,7 @@
 package com.spotify.apollo;
 
 import com.google.auto.value.AutoValue;
+import com.google.common.collect.ImmutableMap;
 
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


rouzwawi:
:+1:

minor comment on `HttpRequest`

pettermahlen:
I will address the ```HttpRequest``` issue in a separate PR, I think it should be removed.

apollo-api/src/main/java/com/spotify/apollo/RequestValue.java
```diff
@@ -57,24 +58,39 @@ private static Request create(
...
-        headers,
+        ImmutableMap.copyOf(parameters),
+        ImmutableMap.copyOf(headers),
         service,
         payload);
   }
 
+  @Override
+  public Request withUri(String uri) {
+    return create(method(), uri, parameters(), headers(), service(), payload());
```
danielnorberg: Should perhaps the parameters be cleared here?

honnix: Rebuild from new URL

Issue #57, opened 2015-12-02
Specify Request.uri more precisely


rouzwawi:
The current javadoc reads:

```
* The uri of the request message.
* Uri query parameters are available via {@link #parameter(String)} and {@link #parameters()}.
* Uri pathArgs are available in {@link RequestContext}.
```

Some details that are unspecified:
- Does the uri contain the scheme of the incoming request?
- Does it contain the authority?
- How does it relate to `Route.uri`?
- What are the differences between an incoming `Request` and one created for an outgoing `Client` call?

pettermahlen:
Some thoughts:

- it would be surprising if the URIs used for outgoing and incoming requests are different.
- the definition of 'URI' includes the scheme (https://tools.ietf.org/html/rfc3986#section-1.1.1).
- there is a concept of 'URI reference', which can be either absolute or relative (https://tools.ietf.org/html/rfc3986#section-4.1). 
- we need the scheme for outgoing URIs, since we might want to use different client implementations to make the calls.

Taken together, this makes me think that we should shoot for absolute URIs. There are probably some backwards compatibility issues if we would do that. 

Another option might be to allow any URI reference for an incoming request and to mandate full URIs for outgoing. I don't like that the two 'kinds' of ```Request``` are different although they have the same type with that solution.

pettermahlen:
I think ```Route.uri``` would be correctly defined as a path fragment. By the time it reaches rut, it's a path (and that's what rut calls it). In ```Route```, we've got the ```withPrefix```. 

I don't think we allow query strings or fragments in Route URIs, right?

pettermahlen:
Adding to 2.0.0 for consideration

pettermahlen:
Too hard for 2.0.0, let's do it later.

lepistone:
Closing because of staleness.

PR #58, opened 2015-12-02
Remove HttpRequest

pettermahlen:
Also defines the 'uri' returned by the Request as including the query string. It's still not well defined whether the 'uri' should include scheme + authority parts, but that's a separate problem (#57).


rouzwawi:
:+1:

PR #60, opened 2015-12-03
Added an example that talks to the Spotify API.

Yarin78:


examples/spotify-api-example/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+    <artifactId>spotify-api-example-service</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+    <packaging>jar</packaging>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.spotify</groupId>
+                <artifactId>apollo-bom</artifactId>
+                <version>1.0.0-SNAPSHOT</version>
```
pettermahlen: Released version? Or the current snapshot.

examples/spotify-api-example/pom.xml
```diff
@@ -0,0 +1,108 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>com.spotify</groupId>
+            <artifactId>apollo-http-service</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-log4j12</artifactId>
```
pettermahlen: why not use logback? It's a nitpick, but it seems like a better choice?

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
```diff
@@ -0,0 +1,109 @@
...
+  private static final String SEARCH_API = SPOTIFY_API + "/v1/search";
+  private static final String ALBUM_API = SPOTIFY_API + "/v1/albums";
+
+  public Stream<Route<AsyncHandler<Response<ByteString>>>> routes() {
+    // The album resource has two routes. Both are using the autoserializer.
+    return Stream.of(
+        Route.async("GET", "/albums/new", this::getAlbums),
+        Route.async("GET", "/albums/hipster", this::getAlbums)
+    )
+        .map(route -> route.withMiddleware(Middlewares::autoSerialize));
```
pettermahlen: I sort of wonder if we should use auto-serialisation in examples. Somebody said that the code you should really make absolutely perfect is example code, because it lives forever. And if we think (like I do) that it's better to be type-safe, then we should use type-safe serialisation.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  public CompletionStage<Response<ArrayList<Album>>> getAlbums(RequestContext requestContext) {
+    String uri = requestContext.request().uri();
+    String tag = uri.substring(uri.lastIndexOf("/") + 1);
```
pettermahlen: The tag should probably be a path parameter instead of parsing it out like this. That is, use

```
 Route.async("GET", "/albums/<tag>", this::getAlbums)
```

as the single route.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * The album resource demonstrates how to use asynchronous routes in Apollo.
+ */
+public class AlbumResource {
+
+  private static final String SPOTIFY_API = "https://api.spotify.com";
+  private static final String SEARCH_API = SPOTIFY_API + "/v1/search";
+  private static final String ALBUM_API = SPOTIFY_API + "/v1/albums";
+
+  public Stream<Route<AsyncHandler<Response<ByteString>>>> routes() {
```
pettermahlen: Since this class has no state, it could be possible to make its methods all static. Not sure if that's better. It's not, because I think the common case is to have some members injected. So... I guess it's OK.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
```diff
@@ -0,0 +1,109 @@
...
+   * Parses an album response from a
+   * <a href="https://developer.spotify.com/web-api/album-endpoints/">Spotify API album query</a>.
+   *
+   * @param json The json response
+   * @return A list of albums with artist information
+   */
+  private ArrayList<Album> parseAlbumData(String json) {
+    ArrayList<Album> albums = new ArrayList<>();
+    try {
+      ObjectMapper mapper = new ObjectMapper();
```
pettermahlen: Should probably make the object mapper into a constant. Or even something that gets injected, just to show that that's possible.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
```diff
@@ -0,0 +1,109 @@
...
+   * Parses the album ids from a JSON response from a
+   * <a href="https://developer.spotify.com/web-api/search-item/">Spotify API search query</a>.
+   *
+   * @param json The JSON response
+   * @return A comma-separated list of album ids from the response
+   */
+  private String parseResponseAlbumIds(String json) {
+    StringJoiner sj = new StringJoiner(",");
+    try {
+      ObjectMapper mapper = new ObjectMapper();
```
pettermahlen: As above.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/ArtistResource.java
```diff
@@ -0,0 +1,113 @@
...
+    return client
+        .send(searchRequest)
+        .thenComposeAsync(response -> {
+          String topArtistId = parseFirstArtistId(response.payload().get().utf8());
+          Request topTracksRequest = Request.forUri(
+              String.format("%s/%s/top-tracks?country=%s", ARTIST_API, topArtistId, country));
+          return client.send(topTracksRequest);
+        }).thenApplyAsync(response -> Response.ok()
+            .withPayload(parseTopTracks(response.payload().get().utf8())))
+        .exceptionally(throwable -> Response
```
pettermahlen: This is unnecessary, since Apollo will be doing that anyway.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/ArtistResource.java
```diff
@@ -0,0 +1,113 @@
...
+   * Parses an artist top tracks response from the
+   * <a href="https://developer.spotify.com/web-api/get-artists-top-tracks/">Spotify API</a>
+   *
+   * @param json The json response
+   * @return A list of top tracks
+   */
+  private ArrayList<Track> parseTopTracks(String json) {
+    ArrayList<Track> tracks = new ArrayList<>();
+    try {
+      ObjectMapper mapper = new ObjectMapper();
```
pettermahlen: Make the object mapper a constant or a constructor parameter.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/ArtistResource.java
```diff
@@ -0,0 +1,113 @@
...
+  /**
+   * Parses the first artist id from a JSON response from a
+   * <a href="https://developer.spotify.com/web-api/search-item/">Spotify API search query</a>.
+   *
+   * @param json The json response
+   * @return The id of the first artist in the response. null if response was empty.
+   */
+  private String parseFirstArtistId(String json) {
+    try {
+      ObjectMapper mapper = new ObjectMapper();
```
pettermahlen: As above

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/SpotifyApiExample.java
```diff
@@ -0,0 +1,46 @@
...
+
+  public static void main(String[] args) throws LoadingException {
+    HttpService.boot(SpotifyApiExample::init, "spotify-api-example-service", args);
+  }
+
+  static void init(Environment environment) {
+    AlbumResource albumResource = new AlbumResource();
+    ArtistResource artistResource = new ArtistResource();
+
+    Stream<Route<AsyncHandler<Response<ByteString>>>> routes =
```
pettermahlen: Should we consider using versioned routes here?

examples/spotify-api-example/pom.xml
Yarin78: Doh, my mistake. Copied this from another example which I guess is wrong as well then.

examples/spotify-api-example/pom.xml
Yarin78: Right, will change.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
Yarin78: Hmm. How would I go about doing that? Should I really have to write one serializer for every domain class? Seems silly. It seems to me this would just increase the chances of introducing bugs. What's the hypthetical case when a mistyping could cause problems? We are converting this to a ByteString after all, which is untyped :)

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
Yarin78: I considered this, but there was two reasons I didn't: 1) To show a resource with multiple routes 2) Because new and hipster are actually the only valid tag values in the Spotify API...

Maybe I could clarify this with a comment, saying that you could do it the other way instead? (parameters are examplified in the other resource)

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
Yarin78: Yes, that was exactly the reason. I don't like classes where everything is static...

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
Yarin78: Hmm. That's seems like a very arbitrary object to inject, why would you want to do that? I admit I hadn't considered testing aspects in this example (probably should be some tests), but not sure how injecting the ObjectMapper (only) would help?

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/ArtistResource.java
Yarin78: Yes, I noticed. It was mainly to show how to handle exceptions.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/ArtistResource.java
Yarin78: Same comment as above, why? (I might have misunderstood some concept here, I just want to parse some json the easiest way possible)

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/SpotifyApiExample.java
Yarin78: For some reason I thought versioned routes was not in open source Apollo. Maybe because it isn't mentioned in the docs :)

Will consider this.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
jpettersson: I typically re-use the same AutoMapper instance across a whole app. I configure it globally and inject it where needed. Is there any advantage to instantiating a new instance on every method call?

Yarin78: I just thought it was a very light weight object without any state so instanciating in reality didn't do anything, and then I just wanted to avoid passing parameters around. But okay, I see that it seems like a best practice to have a single instance of this so I've changed this now.

jsimonsson:
If you want to be cool then add support for access token header in the call! Something that forwards the incoming request Authorization header to the upstream call to api.spotify.com.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/ArtistResource.java
jsimonsson: ObjectMapper is thread safe and you don't need to make a new one every time :) Ie; the new call for each parsing is not needed. You can just have a static final instance somewhere that you reuse.

Yarin78:
Right now I'm getting a scope client, which according to the documentation will have the same context as the incoming request. Not sure if that's what you're asking? 

I could add another route that passes this along for sure. Am I understanding right that you to this service would pass an OAuth token that would get passed along?

jsimonsson:
If all incoming request headers are automatically transfered to an outgoing requests then I'd be very concerned as it is "magic" that happens, especially if it is things such as the "Authorization" header that contains credentials you do not want to leak by accident. I suppose it is fine if all requests made with the client is for "internal" services, but your example is actually the exact opposite, you are calling a "3rd party" service on the interwebz. In your case you want to maybe act as a proxy so you do it by actively transferring the token from incoming -> outgoing.  Maybe someone that implemented that part could tell us :) It was just a suggestion to get some "stuff" into the example, def not needed.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
zalenski: ```.withMiddleware(JsonSerializerMiddlewares.<YourObjectType>jsonSerialize(jackson ObjectWriter))```
will return a type-safe
`Middleware<AsyncHandler<T>, AsyncHandler<Response<ByteString>>>`
for any given `T`

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
zalenski: or pass in "new" or "hipster" as a parameter with
```context -> getAlbums(context, "hipstew")```
to avoid the parsing

**Comment by Yarin78** [comment]
@jsimonsson I checked with Zalenski and apparently that's not how it works, thanksfully :)

I've fixed all (I think) comments @pettermahlen  had on this, and I want to leave this example now. I'd rather add one or two more examples showing other facets of Apollo than building some uberexample. @rouzwawi  could you have a new look at this PR?

File: apollo-api-impl/src/main/java/com/spotify/apollo/route/Middlewares.java [hunk 12]
```diff
@@ -50,7 +50,10 @@ private Middlewares() {
    * into a {@code Response<ByteString>}.
    */
   public static <T> AsyncHandler<Response<ByteString>> autoSerialize(AsyncHandler<T> inner) {
-    return serialize(new AutoSerializer()).apply(inner);
+    Middleware<AsyncHandler<?>, AsyncHandler<Response<ByteString>>>
+        serialize =
```
Yarin78: Oops, this shouldn't have changed.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
Yarin78: Ah, that's nicer of course.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/SpotifyApiExample.java
```diff
@@ -0,0 +1,49 @@
...
+    AlbumResource albumResource = new AlbumResource(objectMapper);
+    ArtistResource artistResource = new ArtistResource(objectMapper);
+
+    Stream<Route<AsyncHandler<Response<ByteString>>>> routes =
+        Stream.of(
+            albumResource.routes(),
+            artistResource.routes(),
+            Stream.of(
+                Route.sync("GET", "/ping", SpotifyApiExample::ping)
+            )
```
rouzwawi: These would be cleaner if registered directly with the routing engine like

```
environment.routingEngine()
    .registerRoutes(albumResource)
    .registerRoutes(artistResource)
    .registerRoute(...ping);
```

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/AlbumResource.java
```diff
@@ -0,0 +1,113 @@
...
+    // Note that this could also have been set up as a single route to "/albums/<tag>".
+    return Stream.of(
+        Route.async("GET", "/albums/new", context -> getAlbums(context, "new")),
+        Route.async("GET", "/albums/hipster", context -> getAlbums(context, "hipster"))
+    )
+        .map(route -> route.withMiddleware(
+            JsonSerializerMiddlewares.jsonSerializeResponse(objectWriter)));
+  }
+
+  public CompletionStage<Response<ArrayList<Album>>> getAlbums(RequestContext requestContext, String tag) {
```
pettermahlen: This could have lower accessibility, since it's only going to be called internally from this class. I think package-private would make sense, so that it could be unit tested.

examples/spotify-api-example/src/main/java/com/spotify/apollo/example/data/Track.java
```diff
@@ -0,0 +1,19 @@
+package com.spotify.apollo.example.data;
```
pettermahlen: I think we should use AutoValue for these classes, or else, at least make them immutable (final fields). AutoValue is much nicer (because you get equals() and toString() for free) if we want to write some tests for this example, which I think we should do.

pettermahlen:
I think this looks good, but we should also add some tests, because:

- Somebody once wrote that example code is the most important code in any product, since people copy/paste it and anything broken in examples lives forever. Therefore, the recommendation was to spend 1x 'normal' time on internal code, 2x 'normal' time on APIs, and 10x 'normal' time on examples. That made sense to me.
- It's a great example for showcasing some of Apollo's nicer test features.

pettermahlen:
If you like, we could merge this, and I could create some tests.

Yarin78:
@pettermahlen I've done some code cleaning up now.

I agree that we should have tests. I'll let you add them, but let me merge this thing now please :)

Regarding AutoValue - I would prefer not to include magic like that in simple examples as it makes the code harder to read (I think) if you're not used to it. For these simple classes, having an explicit toString/equals method is preferable I think. I made the fields final at least for now.

pettermahlen:
:+1:

PR #64, opened 2015-12-07
Clarify metadata interface

### LINKED ISSUE #13: MetaGatherer.EndpointGatherer tracks single content type
This is broken in a few ways:

- the field never gets set
- an endpoint may return different content types for different requests

### LINKED ISSUE #63: Clarify requirements on and behaviour of MetaModule
Some things that are currently more or less broken:

- #13 
- ```MetaModule``` specifies a 'container version' which is hard-coded to 'apollo-standalone' + the apollo version. The latter is retrieved from a file named 'metaDescriptor.properties' on the class path.
- The 'meta' API is exposed on the same port as the main service, under a 'path tree' starting with '_meta'. 
- The APIs are defined in apollo-api-impl, which is not 'public', making it unclear if we're breaking backwards compatibility by changing something like MetaGatherer.

pettermahlen:
This is a set of related changes to improve the metadata collection and presentation. Highlights:

- Move the meta-related interfaces and classes into their own module (except one, IncomingCallsGatherer, which is still in apollo-api-impl).
- Fix #13 - don't track content type at all.
- Fix #63 - make the assembly name used configurable.
- Make it possible for assemblies to define a sort order for client decorators rather than leaving it random. This is necessary to guarantee that outgoing calls will be gathered (if the OutgoingCallsDecorator would end up after the the HttpClientDecorator, it would never be executed).


apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelper.java
```diff
@@ -478,7 +480,18 @@ private void start(int timeoutSeconds, final String... args) throws InterruptedE
             .withModule(HttpClientModule.create())
             .withModule(
                 ForwardingStubClientModule
-                    .create(forwardNonStubbedRequests, stubClient.asRequestAwareClient()));
+                    .create(forwardNonStubbedRequests, stubClient.asRequestAwareClient()))
+            .withModule(new AbstractApolloModule() {
```
rouzwawi: This should probably not be inline like this.

Maybe make the `MetaModule` public with some api for creating it with the assembly name?

pettermahlen: Done - but I'm starting to feel more and more that we should separate out the Meta module from the environment module. A problem with that is that ApolloEnvironmentImpl initialises the metadata collection. But that would probably be OK - it means that ApolloEnvironmentModule requires MetaModule to work, and any assembly using it has to also use the MetaModule?

pettermahlen:
A side note: plus points to japicmp for telling me that I had inadvertently removed the default constructor of ApolloEnvironmentModule in a version of this PR! :)

rouzwawi:
Moving conversation out of commit comment:

> @petter: I'm starting to feel more and more that we should separate out the Meta module from the environment module. A problem with that is that ApolloEnvironmentImpl initialises the metadata collection. But that would probably be OK - it means that ApolloEnvironmentModule requires MetaModule to work, and any assembly using it has to also use the MetaModule?

I'm not sure if the meta module is mature enough to stand on it's own yet. I'm fine if it's still more of an internal thing for now. The usage is tied to the internals of the EnvironmentModule.

If would make more sense to me to make the MetaModule a public module if we can invert the dependency between EnvironmentModule and MetaModule. That might require some added extension points in the EnvironmentModule, but it would make metadata extraction more well defined.

pettermahlen:
@zalenski @rouzwawi what do you think?

apollo-meta/src/main/java/com/spotify/apollo/meta/MetaModule.java
```diff
@@ -17,28 +17,41 @@
...
+  private final String assemblyName;
+
+  public MetaModule(String assemblyName) {
+    this.assemblyName = checkNotNull(assemblyName);
+  }
 
   @Override
   protected void configure() {
+    Multibinder.newSetBinder(binder(), ClientDecorator.class)
+        .addBinding().to(OutgoingCallsDecorator.class);
```
rouzwawi: Does this syntax cause Guice to use the provider method declared below?

apollo-meta/src/main/java/com/spotify/apollo/meta/MetaModule.java
```diff
@@ -17,28 +17,41 @@
...
  * Module for setting up service metadata collection objects.
  */
-class MetaModule extends AbstractModule {
+public class MetaModule extends AbstractApolloModule {
+
+  public static final Id OUTGOING_CALLS = Id.of(MetaModule.class, "Outgoing calls decorator");
+
+  private final String assemblyName;
+
+  public MetaModule(String assemblyName) {
```
rouzwawi: Use static constructor method rather than exposed constructor?

All other modules are creates as `BlaModule.create(...)`

apollo-http-service/src/main/java/com/spotify/apollo/httpservice/HttpServiceModule.java
```diff
@@ -44,14 +53,23 @@ public static HttpServiceModule create(AppInit appInit) {
...
   }
 
   private void bindAppInit() {
     bind(Initializer.class).toInstance(env -> env.initialize(appInit));
   }
 
+  private void bindClientDecoratorComparator() {
+    // ensure that clients track outgoing calls first in the chain
+    bind(new TypeLiteral<Comparator<ClientDecorator.Id>>() { })
+        .toInstance(new ListBasedComparator(singletonList(MetaModule.OUTGOING_CALLS))
```
rouzwawi: Does it make sense to define an ordering when there's only one decorator?

apollo-environment/pom.xml
```diff
@@ -35,6 +35,10 @@
             <artifactId>apollo-api-impl</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.spotify</groupId>
+            <artifactId>apollo-meta</artifactId>
```
rouzwawi: Why does apollo-environment need a dependency to the meta module?

rouzwawi: I see now. The `MetaInfoTracker` is mounted in `ApolloEnvironmentModule`.

rouzwawi:
A few comments:

- `MetaModule` is required if using `ApolloEnvironmentModule`, so the assembly author has to include it when building the `Service`. This is evident in every place where the `ApolloEnvironmentModule` is used.  Yet, it only lets the user specify the assembly name. Can't the assembly name just be an argument to `ApolloEnvironmentModule.create(String assemblyName)` and then having the `MetaModule` be installed internally as it was before? Until `ApolloEnvironmentModule` can work without `MetaModule`, it should own the assembly name.

- The api for specifying client decorator ordering is a Guice dependency, why? Can't also not just be part of different constructors of `ApolloEnvironmentModule.create(...)`, one overload taking `Comparator<ClientDecorator.Id>`, and another taking `List<ClientDecorator.Id>`.

Both these comments are about reducing two aspects of the modular setup:

1. Less invisible hard requirements between modules that are only detected when using one module and seeing that it is missing some dependency that another module provides (in runtime).
2. Less need for users and assembly authors to implement modules for things that essentially are configuration for other modules. These things can simply be constructor arguments to the concerned module.

apollo-meta/pom.xml
```diff
@@ -0,0 +1,147 @@
...
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.spotify</groupId>
+            <artifactId>apollo-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.spotify</groupId>
+            <artifactId>apollo-api-impl</artifactId>
```
rouzwawi: This dependency should go away right? Maybe not in this PR, but I'm thinking that's some sort of end goal we have?

apollo-environment/pom.xml
```diff
@@ -76,6 +85,19 @@
...
+            <artifactId>jmh-core</artifactId>
+            <version>1.3</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.3</version>
+            <scope>test</scope>
+        </dependency>
```
rouzwawi: Should these go into the meta module?

apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelper.java
```diff
@@ -475,10 +476,12 @@ private void start(int timeoutSeconds, final String... args) throws InterruptedE
         Service.Builder serviceBuilder = Services.usingName(serviceName)
             .usingModuleDiscovery(false)
             .withModule(ApolloEnvironmentModule.create())
+            .withModule(new MetaModule("service-helper"))
             .withModule(HttpClientModule.create())
             .withModule(
                 ForwardingStubClientModule
-                    .create(forwardNonStubbedRequests, stubClient.asRequestAwareClient()));
+                    .create(forwardNonStubbedRequests, stubClient.asRequestAwareClient()))
+            .withModule(new ServiceHelperClientDecoratorOrdering());
```
rouzwawi: The api for client decorator order I have in mind would have this look like:

```
Services.usingName(serviceName)
    .usingModuleDiscovery(false)
    .withModule(HttpClientModule.create())
    .withModule(
        ForwardingStubClientModule
            .create(forwardNonStubbedRequests, stubClient.asRequestAwareClient()))
    .withModule(ApolloEnvironmentModule.create("service-helper", OUTGOING_CALLS, STUB_CLIENT));
```

apollo-meta/src/main/java/com/spotify/apollo/meta/MetaModule.java
pettermahlen: Not quite all other modules, but I agree anyway. :) Will fix.

apollo-http-service/src/main/java/com/spotify/apollo/httpservice/HttpServiceModule.java
pettermahlen: There may and often will be more than one. What this means is that 'apply the outgoing calls decorator first, then any others in random order'.

apollo-environment/pom.xml
pettermahlen: They are there, but I forgot to remove them from apollo-environment - now done, thanks!

apollo-http-service/src/main/java/com/spotify/apollo/httpservice/HttpServiceModule.java
rouzwawi: Aha, thanks for clarifying.

pettermahlen:
About module dependencies, I think that ```ApolloEnvironmentModule``` has a dependency on ```apollo-meta```, but not on ```MetaModule```. That is, it requires a ```MetaInfoTracker```. I think the ability to switch out the implementation of the ```MetaInfoTracker``` is useful, and therefore that ```ApolloEnvironmentModule``` shouldn't do the instantiation. It's true that it's less convenient and that errors are only reported at runtime, but then, this is not code that changes often, so for me, the ability to switch out the implementation is worth it.

Having said that, it feels like ```MetaInfoTracker``` should be an interface...?

Changing to an explicit parameter rather than a Guice dependency seems like a great idea. Your suggested API seems to indicate that all decorators must be listed. The current solution doesn't have to be aware of all the decorators, which is good. On the other hand, it can only sort the first few, which is not good. I've been thinking it might be a good idea to have something like

```
   DecoratorOrdering.first(OUTGOING_CALLS, SERVICE_SETTING).last(HERMES, HTTP);
```

That would mean you would get unspecified stuff 'somewhere in the middle'. Maybe it's better to require explicitly defining the order for all decorators; any runtime exceptions could be pretty clearly reported, and you don't often create assemblies, so it's not something people are going to waste a lot of time on.

apollo-meta/pom.xml
pettermahlen: Yes, agreed.

rouzwawi:
> Your suggested API seems to indicate that all decorators must be listed.

Hmm, I didn't intend to suggest that. It would still behave as the current implementation, just get the comparator/list explicitly on module creation.

Having the possibility to specify a "last" ordering seems like a valuable feature though.

rouzwawi:
Also, I agree on meta module reasoning, including MetaInfoTracker being an interface.

pettermahlen:
@rouzwawi updated - yes/no/more improvements?

apollo-environment/src/main/java/com/spotify/apollo/environment/ApolloEnvironmentModule.java
```diff
@@ -65,8 +65,14 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(ApolloEnvironmentModule.class);
 
-  public static ApolloEnvironmentModule create() {
```
rouzwawi: Default creator was removed. It should still be left but cause the default "random" ordering behaviour? Or are we making this a breaking change?

rouzwawi: Reading the other commits, it looks like the idea is to always use `ClientDecoratorOrder` when creating `ApolloEnvironmentModule`.

I would still keep the default, argument-less constructor for simple use cases where the is no need to control the order.

pettermahlen: Yes, this is an API-breaking change. Not in a big way, but yes.

Not specifying an order can be done like so with the current API:
```
  ApolloEnvironmentModule.create((l, r) -> 0);
```

Not a big deal, but I kind of think it's nice to not pollute the API with two constructor methods - I don't know that we have a use case where we don't care about the decorator ordering? At least, we care about it in the currently known ones (HTTP service, Hermes service, and ServiceHelper).

rouzwawi: I've written (and others i know of) command line tools that use the environment module to get a configured Apollo Client. But now as I'm writing this, I realise I could just use the client modules directly.

So it's probably ok :)

rouzwawi:
👍

PR #65, opened 2015-12-09
Per request TTL

gilles:
Allow per request TTL for client calls

### DIFF
--- a/apollo-api/src/main/java/com/spotify/apollo/Request.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/Request.java
@@ -88,7 +88,7 @@ public interface Request {
   Optional<ByteString> payload();
 
   /**
-   * The request ttl in ms
+   * The request ttl.
    */
   default Optional<Duration> ttl() {
     return Optional.empty();
@@ -144,19 +144,10 @@ public interface Request {
   /**
    * Creates a new {@link Request} based on this, but with a different ttl.
    *
-   * @param ttl The duration, will be turned into ms
+   * @param ttl The duration
    */
   default Request withTtl(final Duration ttl) {
     throw new UnsupportedOperationException();
-  }
-
-  /**
-   * Creates a new {@link Request} based on this, but with a different ttl in ms
-   *
-   * @param ttl The duration in ms
-   */
-  default Request withTtl(final int ttl) {
-    return withTtl(Duration.ofMillis(ttl));
   }
 
   /**
--- a/apollo-api/src/main/java/com/spotify/apollo/Request.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/Request.java
@@ -22,6 +22,7 @@ package com.spotify.apollo;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.TimeUnit;
 
 import okio.ByteString;
 
@@ -87,6 +88,11 @@ public interface Request {
   Optional<ByteString> payload();
 
   /**
+   * The request ttl in ms
+   */
+  Optional<Long> ttl();
+
+  /**
    * Creates a new {@link Request} based on this, but with a different URI.
    *
    * @param uri the new uri
@@ -132,6 +138,21 @@ public interface Request {
    * @param payload the new payload
    */
   Request withPayload(ByteString payload);
+
+  /**
+   * Creates a new {@link Request} based on this, but with a different ttl in ms.
+   *
+   * @param ttl The ttl in ms
+   */
+  Request withTtl(long ttl);
+
+  /**
+   * Creates a new {@link Request} based on this, but with a different ttl in ms.
+   *
+   * @param ttl      The ttl
+   * @param timeUnit The timeunit of the ttl
+   */
+  Request withTtl(long ttl, TimeUnit timeUnit);
 
   /**
    * Creates a {@link Request} for the given uri and method.
--- a/apollo-api/src/main/java/com/spotify/apollo/Request.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/Request.java
@@ -19,10 +19,10 @@
  */
 package com.spotify.apollo;
 
+import java.time.Duration;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.TimeUnit;
 
 import okio.ByteString;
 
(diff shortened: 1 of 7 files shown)


pettermahlen:
Could you describe the rationale for this change, please? What are the things you want to achieve (I can guess from reading the code, but I might be wrong, and it's nice to have it documented here)?

pettermahlen:
Here are a couple of other comments, based on a discussion with @rouzwawi:

First, on the API:
- We think that adding the concept of TTL is a good thing.
- We'd prefer using the `java.time.Duration` type in the interface rather than a Long or Long/TimeUnit combination. It's more modern and also helps make things symmetrical when setting/getting.
- The change as currently written breaks backwards compatibility. The getter and mutator methods should have default implementations to prevent that. Probably, the getter should return empty() and the mutator should throw UnsupportedOperationException.

Second, on the HTTP client implementation:
- The timeout is never set on the cloned OkHttpClient, so the tests are failing to catch that.
- The tests are tightly tied to the implementation (invoking specific methods on OkHttpClient) as opposed to validating the behaviour (verifying that requests in fact time out as per the specified limit).

Does that make sense to you?

gilles:
And I learned about `Duration` :) 

If I understand well you'd like
```java
Optional<Duration> ttl()
Request withTtl(Duration duration)
```

The compatibility breaks if you use apollo-api = 1.0.0 and apollo-okhttp-client > 1.0.0, which seems an odd case as we provide the BOM. And I'm not sure how providing defaults implementations (living in apollo-api) would prevent the failure. I'm not sure I follow actually...

I haven't found how to make `MockServerRule` wait and simulate a timeout so I went with mockito.

gilles:
I think I just find out how to simulate a timeout, will try

pettermahlen:
The compatibility issue I was referring to is the breaking change in the Request interface. It can be solved like so in ```Request```:

```
  default Optional<Duration> ttl() {
     return empty();
  }

  default Request withTtl(Duration ttl) {
    throw new UnsupportedOperationException();  
  }
```

Then in ```RequestValue```, there would be actual implementations of the functionality.

gilles:
`AutoValue` will add a `ttl()`, from the generated code:

```
@Override
  public Optional<Duration> ttl() {
    return ttl;
}
```

And `withTtl()` is defined in `RequestValue` by me. Since `Request`, `RequestValue` and `AutoValue_RequestValue` are in the same module I'm still unclear what could be the issue (sorry for being thick)

gilles:
latest failure seems unrelated and it all tests pass locally, is there a way to retrigger?

rouzwawi:
You can retrigger on CricleCI I believe. I did it now.

On the incompatibility issues: `Request` is implemented outside of this code-base as well, and those implementations would break when the new interface methods don't have default implementations. This can lead to runtime `MethodNotFoundException`.

gilles:
haaaa
>> Request is implemented outside of this code-base as well
I see what you mean now. 

I'll provide the `default`

gilles:
There's an issue with `default Optional<Duration> ttl()` as `AutoValue` does not like it -> the `ttl` field does not end up in `RequestValue`  and there is no implementation for `ttl()` there so there's no way to set it / retrieve it anymore

And you can't do (won't compile)
```
Optional<Duration> ttl();
default Optional<Duration> ttl() {};
```

Not quite sure how to fix this.

pettermahlen:
Looks like AutoValueProcessor doesn't override methods that have the modified 'default': https://github.com/google/auto/blob/master/value/src/main/java/com/google/auto/value/processor/AutoValueProcessor.java#L642. I can see a couple of possible ways to work around that:

1. Add ```public abstract Optional<Duration> ttl()``` to ```RequestValue```.
2. Don't use AutoValue for ```RequestValue```. 

I did a quick test of option 1, and it seems to work from what I can tell. A little weird, but it may be the easiest path forward.

apollo-api/src/main/java/com/spotify/apollo/Request.java
```diff
@@ -134,6 +142,24 @@
...
+  default Request withTtl(final Duration ttl) {
+    throw new UnsupportedOperationException();
+  }
+
+  /**
+   * Creates a new {@link Request} based on this, but with a different ttl in ms
+   *
+   * @param ttl The duration in ms
+   */
+  default Request withTtl(final int ttl) {
```
pettermahlen: I don't think we should have this - it's better to have as few methods as possible, and in my opinion, this doesn't carry its weight.

apollo-api/src/main/java/com/spotify/apollo/Request.java
```diff
@@ -134,6 +142,24 @@
   Request withPayload(ByteString payload);
 
   /**
+   * Creates a new {@link Request} based on this, but with a different ttl.
+   *
+   * @param ttl The duration, will be turned into ms
```
pettermahlen: 'turned into ms' doesn't seem applicable any more?

apollo-api/src/main/java/com/spotify/apollo/Request.java
```diff
@@ -87,6 +88,13 @@
   Optional<ByteString> payload();
 
   /**
+   * The request ttl in ms
```
pettermahlen: Not 'ms' any more?

apollo-api/src/main/java/com/spotify/apollo/RequestValue.java
```diff
@@ -55,23 +56,27 @@ private static Request create(
...
         method, uri,
         ImmutableMap.copyOf(parameters),
         ImmutableMap.copyOf(headers),
         service,
-        payload);
+        payload,
+        ttl);
   }
 
+  public abstract Optional<Duration> ttl();
```
pettermahlen: Should have an ```@Override``` annotation to ensure that it doesn't accidentally become an additional parameter.

apollo-api/src/main/java/com/spotify/apollo/Request.java
```diff
@@ -134,6 +142,15 @@
   Request withPayload(ByteString payload);
 
   /**
+   * Creates a new {@link Request} based on this, but with a different ttl.
+   *
+   * @param ttl The duration
+   */
+  default Request withTtl(final Duration ttl) {
```
rouzwawi: Don't specify `final`, just as the other `with*` methods.

rouzwawi:
LGTM :+1:

just a note on the `final` keyword on the interface method.

rouzwawi:
Great, thanks!

PR #66, opened 2015-12-16
[apollo-core] add option to overlay config with explicit file

rouzwawi:
@petter @udoprog


pettermahlen:
LGTM except for the location of the 'overlay.conf' file. That should go into something like ```src/test/non-classpath-resources```. With a better name probably. :)

pettermahlen:
:+1:

PR #67, opened 2015-12-17
improve test of http client timeouts

pettermahlen:
The previous didn't fail if no SocketTimeoutException was thrown. This one does, and I also changed a couple of constants so as to reduce the test run time a bit.


gilles:
nice
:+1:

rouzwawi:
:+1:

PR #68, opened 2015-12-17
don't include mockito in compile scope

pettermahlen:


gilles:
Actually we should not need mockito anymore since we use the callback in mockserver to simulate the provoke the timeout.

If I remember well I added it in the first version of my tests.

Side note, instead of setting `<scope>test</scope>` in the module poms, should that be in the parent's (apollo-bom) `<dependencyManagement>` section?

PR #78, opened 2016-02-02
Update DEPLOYING.md

zalenski:
@pettermahlen

### DIFF
--- a/DEPLOYING.md
+++ b/DEPLOYING.md
@@ -13,9 +13,9 @@ Once you've got that in place, you should be able to do deployment using the fol
 
 ```
 # deploy snapshot version
-mvn clean deploy
+mvn clean deploy # -Prelease to test signing
 
-# make and deploy a relase
+# make and deploy a release
 mvn release:clean release:prepare -Prelease
-mvn release:perform
+mvn release:perform -Prelease
 ```


DEPLOYING.md
```diff
@@ -13,9 +13,9 @@ Once you've got that in place, you should be able to do deployment using the fol
 
 ```
 # deploy snapshot version
-mvn clean deploy
+mvn clean deploy # -Prelease to test signing
 
-# make and deploy a relase
+# make and deploy a release
```
zalenski: spelling

DEPLOYING.md
```diff
@@ -13,9 +13,9 @@ Once you've got that in place, you should be able to do deployment using the fol
 
 ```
 # deploy snapshot version
-mvn clean deploy
+mvn clean deploy # -Prelease to test signing
```
zalenski: -Prelease to test signing?

DEPLOYING.md
```diff
@@ -13,9 +13,9 @@ Once you've got that in place, you should be able to do deployment using the fol
...
 
-# make and deploy a relase
+# make and deploy a release
 mvn release:clean release:prepare -Prelease
-mvn release:perform
+mvn release:perform -Prelease
```
zalenski: needed?

rouzwawi: This activates the staging plugin, so probably needed.

pettermahlen:
:+1:

pettermahlen:
I believed that we should also move the release plugin config outside the 'release' profile, so you wouldn't have to specify ```-Prelease``` at all. But from what I can tell, the ```releaseProfiles``` setting is only applied to the 'perform' goal. That might mean that -Prelease isn't necessary for 'perform', but I think better safe than sorry..

rouzwawi:
I'm ok with moving it out if it doesn't effect normal builds. `mvn verify`

pettermahlen:
I wasn't clear in my previous comment. What I meant was, "I used to think it was a good idea to move it out, but I've changed my mind". So I think this update to the instructions is good, and we don't need to make other changes.

rouzwawi:
@pettermahlen ah I misread

PR #79, opened 2016-02-03
add arrival timestamp to `RequestContext`

danielnorberg:


danielnorberg:
Where is a good place to put in a unit test for this?

rouzwawi:
The `RequestContext` is created pretty late in the chain. Perhaps you want to take the time of when the `OngoingRequest` is created, which is as soon as the the receiving server starts handling the request.

`RequestHandlerImplTest` might be a good place to test it. But both `OngoingRequest` and `RequestContext` are pretty much just value types. Not too much there to test.

rouzwawi:
:+1:

PR #84, opened 2016-02-21
corrected typo

geoaxis:
microservices should not be spelled with a "-".

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 Apollo
 ======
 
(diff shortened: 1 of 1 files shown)


rouzwawi:
The most commonly use spelling is "microservices". The Wikipedia article uses this spelling as well: https://en.m.wikipedia.org/wiki/Microservices

rouzwawi:
Sorry, I interpreted this PR the other way around.
Thanks for the fix.

PR #86, opened 2016-02-24
configurable incoming request reaping interval

danielnorberg:
Allow specifying desired incoming request reaping interval using a system property.

E.g. `-Dcom.spotify.apollo.request.reaping_interval_ms=1000`.

This is the first step towards potentially removing it completely and thus I assume it is sufficient to only have this be configurable using a system property.

### DIFF
--- a/apollo-api-impl/src/main/java/com/spotify/apollo/request/RequestTracker.java
+++ b/apollo-api-impl/src/main/java/com/spotify/apollo/request/RequestTracker.java
@@ -51,6 +51,8 @@ public class RequestTracker implements Closeable {
               .setNameFormat("apollo-request-reaper")
               .build());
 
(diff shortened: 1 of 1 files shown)


danielnorberg:
@pettermahlen @zalenski PTAL

rouzwawi:
I'd like to know more about removing it and the steps you had in mind.

Even though request reaping and tracking ongoing requests should be managed by the server library, it's useful to have an Apollo controlled facility that is server agnostic.

As to making it configurable, setting the value based on loaded Config can easily be done [here](https://github.com/spotify/apollo/blob/98b577b2bc23239323f896d4c992e25d90a7c9de/apollo-environment/src/main/java/com/spotify/apollo/environment/ApolloEnvironmentModule.java#L187). The loaded config will contain system properties.

pettermahlen:
I guess this is no longer really relevant since #87 ?

danielnorberg:
Yeah, closing this.

PR #87, opened 2016-02-24
remove redundant request tracking

danielnorberg:
Both jetty and hermes servers handle request timeouts so there is no need for apollo to also perform incoming request timeout tracking.

### DIFF
--- a/apollo-api-impl/src/main/java/com/spotify/apollo/request/Handlers.java
+++ b/apollo-api-impl/src/main/java/com/spotify/apollo/request/Handlers.java
@@ -54,6 +54,7 @@ public final class Handlers {
     return new EndpointInvocationHandler();
   }
 
+  @Deprecated
   public static EndpointRunnableFactory withTracking(
       EndpointRunnableFactory endpointRunnableFactory,
       IncomingCallsGatherer incomingCallsGatherer,
@@ -63,5 +64,13 @@ public final class Handlers {
         endpointRunnableFactory,
         incomingCallsGatherer,
         requestTracker);
+  }
+
+  public static EndpointRunnableFactory withGathering(
+      final EndpointRunnableFactory endpointRunnableFactory,
+      final IncomingCallsGatherer incomingCallsGatherer) {
+    return new GatheringEndpointRunnableFactory(
+        endpointRunnableFactory,
+        incomingCallsGatherer);
   }
 }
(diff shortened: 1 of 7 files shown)


apollo-api-impl/src/main/java/com/spotify/apollo/request/TrackingEndpointRunnableFactory.java
```diff
@@ -26,6 +26,7 @@
 /**
  * An {@link EndpointRunnableFactory} that collect statistics
  */
+@Deprecated
```
rouzwawi: This class can be removed since it's not used any more.

apollo-api-impl/src/main/java/com/spotify/apollo/request/Handlers.java
```diff
@@ -64,4 +65,11 @@ public static EndpointRunnableFactory withTracking(
         incomingCallsGatherer,
         requestTracker);
   }
+
+  public static EndpointRunnableFactory withGathering(final EndpointRunnableFactory endpointRunnableFactory,
```
rouzwawi: formatting as above?

rouzwawi:
some minor comments. but :+1: otherwise

apollo-api-impl/src/main/java/com/spotify/apollo/request/TrackingEndpointRunnableFactory.java
danielnorberg: It's still called from [`Handlers#withTracking()`](https://github.com/spotify/apollo/pull/87/files#diff-a9d3e4572f3699295840da558f6d80b6R63) though, so unless we want to break the public api now I guess we need to keep it around?

rouzwawi: Ah, forgot about that.

zalenski:
LGTM

pettermahlen:
:+1: on the code - what validation of timeout handling has been made? Do we have integration tests that verify that requests time out if there is no response? If not, that would be something we should add. (Possibly in another PR if you haven't got the time right now).

danielnorberg:
We know that hermes enforces timeouts and I read the jetty source code to convince myself that jetty does as well.

https://github.com/eclipse/jetty.project/blob/master/jetty-server/src/main/java/org/eclipse/jetty/server/HttpChannelState.java#L899

https://github.com/eclipse/jetty.project/blob/master/jetty-server/src/main/java/org/eclipse/jetty/server/AsyncContextEvent.java#L156

Did not find any existing apollo integration tests that verifies that timeouts are enforced.

PR #89, opened 2016-04-03
Add apollo-entity middleware

rouzwawi:
A set of middleware for working with typed route handlers.

---

The gist of it

```java
class MyApplication {

  private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();

  static void init(Environment environment) {
    MyApplication app = new MyApplication();
    EntityMiddleware entity = EntityMiddleware.forCodec(JacksonEntityCodec.forMapper(OBJECT_MAPPER));

    Route<SyncHandler<Response<ByteString>>> route = Route.with(
        entity.direct(Person.class),
        "PUT", "/person/<id>",
        rc -> person -> app.updatePerson(rc.pathArgs().get("id"), person));

    // ...
  }

  Person updatePerson(String id, Person person) {
    // ...
  }
}
```

There's also support for other handler signatures. See README in PR.

### DIFF
--- a/apollo-entity/README.md
+++ b/apollo-entity/README.md
@@ -127,4 +127,4 @@ See [`EntityTest`][2] for a complete list of route options and tests.
 
 [1]: src/main/java/com/spotify/apollo/entity/EntityMiddleware.java
 [2]: src/test/java/com/spotify/apollo/entity/EntityTest.java
-[project skeleton]: https://github.com/spotify/apollo/tree/master/apollo-http-service#minimal-project-skeleton
+[project skeleton]: https://github.com/spotify/apollo/tree/master/apollo-http-service#maven
--- a/apollo-entity/README.md
+++ b/apollo-entity/README.md
@@ -55,7 +55,7 @@ class MyApplication {
 
   static void init(Environment environment) {
     MyApplication app = new MyApplication();
-    EntityMiddleware entity = EntityMiddleware.jackson(OBJECT_MAPPER);
+    EntityMiddleware entity = EntityMiddleware.forJackson(OBJECT_MAPPER);
 
     Route<SyncHandler<Response<ByteString>>> route = Route.with(
         entity.direct(Person.class),
@@ -70,6 +70,10 @@ class MyApplication {
   }
 }
 ```
+
+[`EntityMiddleware`][1] also support custom codecs through the `forCodec(EntityCodec)` static
+constructor. An [`EntityCodec`][2] defines how to read and write entity types from the `ByteString`
+payloads that apollo handle natively.
 
 ## Handler signatures
 
@@ -127,8 +131,9 @@ rc -> this::handler
 
 ---
 
-See [`EntityTest`][2] for a complete list of route options and tests.
+See [`EntityMiddlewareTest`][3] for a complete list of route options and tests.
 
 [1]: src/main/java/com/spotify/apollo/entity/EntityMiddleware.java
-[2]: src/test/java/com/spotify/apollo/entity/EntityTest.java
+[2]: src/main/java/com/spotify/apollo/entity/EntityCodec.java
+[3]: src/test/java/com/spotify/apollo/entity/EntityMiddlewareTest.java
 [project skeleton]: https://github.com/spotify/apollo/tree/master/apollo-http-service#maven
--- a/apollo-entity/README.md
+++ b/apollo-entity/README.md
@@ -76,14 +76,16 @@ class MyApplication {
 The [`EntityMiddleware`][1] interface can create middlewares for several types of handler
 signatures.
 
-* `direct(E.class)` middleware : `EntityHandler<E, R>`
+* `direct(E.class[, R.class])` middleware : `EntityHandler<E, R>`
  - `R handler(E)`
-* `response(E.class)` middleware : `EntityResponseHandler<E, R>`
+* `response(E.class[, R.class])` middleware : `EntityResponseHandler<E, R>`
  - `Response<R> handler(E)`
-* `asyncDirect(E.class)` middleware : `EntityAsyncHandler<E, R>`
+* `asyncDirect(E.class[, R.class])` middleware : `EntityAsyncHandler<E, R>`
  - `CompletionStage<R> handler(E)`
-* `asyncResponse(E.class)` middleware : `EntityAsyncResponseHandler<E, R>`
+* `asyncResponse(E.class[, R.class])` middleware : `EntityAsyncResponseHandler<E, R>`
   - `CompletionStage<Response<R>> handler(E)`
+
+For all of the above, if the reply entity type `R` is omitted it will be set to the same as `E`.
 
 On top of these, there's also just a serializing middleware for handlers that do not take a
 request entity as an input.
@@ -97,9 +99,9 @@ request entity as an input.
 * `asyncSerializerResponse(R.class)` middleware : `AsyncHandler<Response<R>>`
  - `CompletionStage<Response<R>> handler()`
 
-All of the `Entity*` handler signatures are in curried form with a `RequestContext` as the first
-argument. This give you the flexibility to add more arguments to your handlers as you see fit,
-while also being easy to compose.
+All of the `Entity*Handler<E, R>` handler signatures are in curried form with a `RequestContext`
+as the first argument. This give you the flexibility to add more arguments to your handlers as you
+see fit, while also being easy to compose.
 
 ```java
 rc -> entity -> handler(entity, ...)
--- a/apollo-entity/README.md
+++ b/apollo-entity/README.md
@@ -55,7 +55,7 @@ class MyApplication {
 
   static void init(Environment environment) {
     MyApplication app = new MyApplication();
-    EntityMiddleware entity = EntityMiddleware.forJackson(OBJECT_MAPPER);
+    EntityMiddleware entity = EntityMiddleware.forCodec(JacksonEntityCodec.forMapper(OBJECT_MAPPER));
 
     Route<SyncHandler<Response<ByteString>>> route = Route.with(
         entity.direct(Person.class),
@@ -73,7 +73,8 @@ class MyApplication {
 
 [`EntityMiddleware`][1] also support custom codecs through the `forCodec(EntityCodec)` static
 constructor. An [`EntityCodec`][2] defines how to read and write entity types from the `ByteString`
-payloads that apollo handle natively.
+payloads that apollo handle natively. Since Jackson is so commonly used, we have a bundled
+implementation called [`JacksonEntityCodec`][4].
 
 ## Handler signatures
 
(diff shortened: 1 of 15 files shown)


apollo-entity/src/main/java/com/spotify/apollo/entity/CodecEntityMiddleware.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+import java.util.function.Function;
+
+import okio.ByteString;
+
+import static java.util.concurrent.CompletableFuture.completedFuture;
+
+/**
+ * A factory for creating middlewares that can be used to jackson the routes that deal directly
```
mattnworb: "can be used to jackson"?

rouzwawi: hehe, I have no idea. I should stop editing english as I edit code :P

rouzwawi:
Open question: should the `direct[response]` middleware expect the same entity type on both request and response payloads or let the user specify different types?

apollo-entity/README.md
```diff
@@ -0,0 +1,134 @@
...
+
+```xml
+<dependency>
+    <groupId>com.spotify</groupId>
+    <artifactId>apollo-entity</artifactId>
+    <version>1.0.5</version>
+</dependency>
+```
+
+_See the [project skeleton] docs for how to use a Maven BOM for dependency versions_
```
pettermahlen: Probably fix this link before merge?

rouzwawi: But it's linked to https://github.com/spotify/apollo/tree/master/apollo-http-service#maven which is added in this PR.

apollo-entity/src/main/java/com/spotify/apollo/entity/CodecEntityMiddleware.java
```diff
@@ -0,0 +1,250 @@
...
+  @Override
+  public <E, R> Middleware<EntityAsyncResponseHandler<E, R>, AsyncHandler<Response<ByteString>>>
+  asyncResponse(Class<? extends E> entityClass, Class<? extends R> entityResponseClass) {
+    return inner ->
+        Middleware.syncToAsync(deserialize(entityClass))
+            .flatMap(r -> ctx -> asyncFlatMapPayload(r, inner.invoke(ctx)))
+            .map(serialize(entityResponseClass));
+  }
+
+  private <E> SyncHandler<Response<E>> deserialize(Class<? extends E> entityClass) {
```
pettermahlen: I'm not sure I'm reading all the types right here, but it seems like in this case, we're using a ```Response<E>``` as an input to the real request processing? That feels odd. What is the rationale? ```mapPayload``` taking a ```Response``` as input seems backwards.

rouzwawi: The idea is that we can chain the actual processing on to the `Response<E>` created by the deserialization step. If the deserialization fails, this function return a `Response<E>` with no payload and just a error status code with an explanation. Since everything else is operating on the payload, this will short circuit the request and just return the failed deserialization error.

pettermahlen: I think I get the mechanics, and maybe it's the purist in me that thinks it's wrong to use a ```Response``` as an input/request when the type is defined as a ```Response```. It could be leading to complications to add a type and handle errors based on that type, I suppose. But at least to me, it's surprising and a bit weird to use the ```Response``` type as input to request processing.

rouzwawi: Which parts are you referring to specifically?

I see it as the `Response<E>` that is created in the deserialization step actually is a response to the request, but only in the case when things go wrong. So it's a type encoding the response path. The user code will only ever see the payload of that step, so there will never be any `Response` instance passed out to user code.

As for `mapPayload(response, fn)`, it's just a way for me to do `response.mapPayload(fn)` without adding `mapPayload` to the `Response` interface.

pettermahlen: But in the case that things go right, the payload of the 'response' created during deserialization is in fact the *request* payload, right? That's the thing that looks odd. Calling something a response when it's in fact the request. I was totally confused when reading the deserialize method, because it seemed to me that the codec would have to apply the endpoint handler function and I didn't see how that could work.

I'm more concerned about future maintenance than users. From an API perspective, I really like this PR. It's just that it's hard (for me) to navigate this code.

rouzwawi: Ah I see. Hadn't thought of it that way. I'll see if I can decouple the request entity from the whole response chain in a nice way.

rouzwawi: An idea: what if the deserialize function could take a continuation function for doing the transformation `E -> Response<R>`. Perhaps also be called something else.

rouzwawi: Fixed!

pettermahlen:
Things I like about the last commit:
- the ```withEntityX``` names, which make it clearer to me that the entity is 'added' at that step.
- getting rid of the ```ensureResponse``` method. :) 
- the clarification of what the E and R type parameters mean. I think that could be made even clearer by naming the ```Class<? extends E>``` parameters ```requestEntityClass``` and the ```Class<? extends R>``` parameters ```responseEntityClass```.

Probably need to digest this a little more to see if there are more things I personally struggle with. :)

rouzwawi:
> Coverage decreased (-0.01%) to 68.202%

What, the new code has 100% coverage!

rouzwawi:
> Coverage decreased (-0.01%) to 68.202%

What, the new code has 100% coverage!

rouzwawi:
> Coverage decreased (-0.01%) to 68.202%

What, the new code has 100% coverage!

apollo-entity/src/test/java/com/spotify/apollo/entity/JsonMatchers.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.hamcrest.Description;
+import org.hamcrest.Matcher;
+import org.hamcrest.TypeSafeMatcher;
+
+import okio.ByteString;
+
+import static com.jayway.jsonassert.JsonAssert.with;
+
+class JsonMatchers {
```
gilles: there's a `hamcrest-json` library, though I kind of like the `hasJsonPath` helper.

rouzwawi: Yeah I had a brief look at it and stopped considering it as soon as I saw the first example which showcased inline written json in the asserts :)

```
sameJSONAs("{\"friend_ids\":[52, 23, 16]}")
```

rouzwawi:
1.0.5-SNAPSHOT is pushed to sonatype in case anyone is interested in giving this a try before it's merged.

apollo-entity/src/main/java/com/spotify/apollo/entity/EntityMiddleware.java
```diff
@@ -0,0 +1,113 @@
...
+  <R> Middleware<SyncHandler<Response<R>>, SyncHandler<Response<ByteString>>>
+  serializerResponse(Class<? extends R> responseEntityClass);
+
+  <R> Middleware<AsyncHandler<R>, AsyncHandler<Response<ByteString>>>
+  asyncSerializerDirect(Class<? extends R> responseEntityClass);
+
+  <R> Middleware<AsyncHandler<Response<R>>, AsyncHandler<Response<ByteString>>>
+  asyncSerializerResponse(Class<? extends R> responseEntityClass);
+
+  interface EntityHandler<E, R>
```
pettermahlen: These type signatures are hard to decipher (for me). Is it necessary to have them extend ```SyncHandler``` for them to fit in with the rest of Apollo? I've only thought about this rather than tried it, but I get the feeling it shouldn't be, since the middleware will end up translating it into an ```AsyncHandler<Response<ByteString>>```. If it is not necessary, I would prefer to not extend that type, since doing so signals that these are endpoint handlers in their own right. And they're not, they just happen to have the same signature. Or?

apollo-entity/src/main/java/com/spotify/apollo/entity/EntityCodec.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.IOException;
+
+/**
+ * Interface for serializing and de-serializing entity types.
+ */
+public interface EntityCodec {
+
+  String defaultContentType();
+
+  <E> byte[] write(E entity, Class<? extends E> clazz) throws IOException;
```
pettermahlen: We're using ```okio.ByteString``` everywhere except here, why choose a mutable byte array in this interface?

apollo-entity/src/main/java/com/spotify/apollo/entity/EntityMiddleware.java
rouzwawi: fixed

apollo-entity/src/main/java/com/spotify/apollo/entity/EntityCodec.java
rouzwawi: I had a short thought about not doing unnecessary  wrapping of bytebuffers for performance reasons. But this is not really any better since the api forces the user to allocate a buffer.

I could write this api in terms of input/output streams to make it possible to be more efficient in the middleware later. But there's so many more places in apollo where byte-copying needs to improve.

What do you say, should we start here, or keep it as is and take the cost later (probably apollo 2.0.0)

pettermahlen: I would think it's best to be consistent for now, and that would mean we should use ByteString here, and as you say, perhaps address efficiency concerns in a future version of the API?

pettermahlen:
👍

rouzwawi:
@pettermahlen also fixed the jackson impl thing.

apollo-entity/README.md
```diff
@@ -136,4 +137,5 @@ See [`EntityMiddlewareTest`][3] for a complete list of route options and tests.
 [1]: src/main/java/com/spotify/apollo/entity/EntityMiddleware.java
 [2]: src/main/java/com/spotify/apollo/entity/EntityCodec.java
 [3]: src/test/java/com/spotify/apollo/entity/EntityMiddlewareTest.java
+[4]: src/test/java/com/spotify/apollo/entity/JacksonEntityCodec.java
```
pettermahlen: Should that not be src/main/... ?

rouzwawi: good catch

rouzwawi:
@pettermahlen do you want to have the honours of pressing merge :)

pettermahlen:
I do! (Not that I deserve the honours, but I like this PR. :) )

rouzwawi:
\o/

PR #91, opened 2016-04-07
add protobuf middlewares

rgruener:
@rouzwawi

### DIFF
--- a/apollo-extra/pom.xml
+++ b/apollo-extra/pom.xml
@@ -55,6 +55,12 @@
             <version>2.3.22</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+            <version>3.0.0-beta-2</version>
+            <scope>provided</scope>
+        </dependency>
 
         <!--test scope-->
         <dependency>
--- a/apollo-extra/src/main/java/com/spotify/apollo/route/ProtobufSerializerMiddlewares.java
+++ b/apollo-extra/src/main/java/com/spotify/apollo/route/ProtobufSerializerMiddlewares.java
@@ -1,0 +1,78 @@
+/*
+ * -\-\-
+ * Spotify Apollo Extra
+ * --
+ * Copyright (C) 2013 - 2015 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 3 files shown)


rgruener:
I originally included a proto2json middleware as well however it relied on spotify internals. protobuf3 now includes [printing to json](https://github.com/google/protobuf/blob/master/java/util/src/main/java/com/google/protobuf/util/JsonFormat.java#L175) but I do not think that should be included since many users of protobuf within spotify are not using protobuf3 yet.

rgruener:
Poke

yonromai:
Such wow. @rouzwawi ?

rouzwawi:
LGTM, @pettermahlen thoughts?

@rgruener is the idea with the `provided` dependency and relying on `MessageLite` that this would be backwards compatible with older protobuf versions?

pettermahlen:
I also don't understand the pros and cons of this change (hence the silence; I should have asked instead, sorry). Here are some more specific questions: 

- Adding protobuf as a dependency is historically a scary and unsafe thing to do. What's the rationale, what's the use case for this? 
- Why did you pick 'provided' scope? What are the options? How will users find out that they need to include protobuf into their service builds if they do this?
- If there are risks of backwards incompatibility, could this be done as a separate library outside of apollo-extra?

rgruener:
@rouzwawi Yeah the idea was to make it backwards compatible and not pull in protobuf unless you are using it already. I didn't want to cause enforcer problems where you would need exclusions. Perhaps optional would be more appropriate.

@pettermahlen The rational is that we still use protobuf a lot internally within Spotify and since there is no official middleware for it people are often using deprecated serializers or writing their own custom ones which involves copying a lot of code between projects. Before making this pr I had a brief discussion with Rouz whether this should be in the open source version of apollo and since it is useful for a broader audience than only Spotifiers it could go in apollo-extra.

I understand that pulling in protobuf can be tricky and that was the reasoning behind the provided scope. If your service has protobuf payloads it should include protobuf already since I never really thought of apollo-extra as something that should be . Provided is a little sneaky though perhaps optional is more fitting. I don't have such strong opinions on if this is great to place in apollo-extra but it would definitely be useful to be pulled in by default for Spotify apollo services.

pettermahlen:
Reading this, https://maven.apache.org/guides/introduction/introduction-to-optional-and-excludes-dependencies.html, it sounds like the recommended solution is to put it into a separate artifact, if possible.

When you say, 'pulled in by default', what do you mean? I guess protobuf shouldn't be pulled in by default, for all the compatibility issues it causes, but I think it makes sense to make it easy for people to do protobuf serialization correctly.

I'm sort of leaning towards the best solution being to have an apollo-protobuf maven module (not apollo or guice module).

rgruener:
I think a separate maven module is a good solution. Does it belong in this repo?

pettermahlen:
@rgruener i would say so, yes.

yonromai:
@rgruener any news regarding this PR?

rgruener:
yeah closing this

Issue #71, opened 2015-12-18
Request#header(String) should be case insensitive for HTTP requests


udoprog:
See http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html
From 4.2 Message Headers: "Field names are case-insensitive."

This will otherwise cause problems when interacting with proxies or clients that perform pre-emptive all-to-lowercase/uppercase [example here](https://github.com/nodejs/node-v0.x-archive/issues/1954) before the request hits the Apollo-based service.

pettermahlen:
What exactly is the problem in the Apollo code base?

udoprog:
[Request#header(String)](https://github.com/spotify/apollo/blob/master/apollo-api/src/main/java/com/spotify/apollo/Request.java#L77) is not case-insensitive, which should probably happen in [ApolloRequestHandler in the jetty-http-server module](https://github.com/spotify/apollo/blob/master/modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java#L90)

pettermahlen:
Right, of course, it was in the title. We should define whether ```Request``` itself considers headers to be case-sensitive or not, I think. I don't have any strong opinions about that, but I think I would like to make it case-insensitive like for HTTP.

pettermahlen:
Addressed in #93

pettermahlen:
Fixed via #96

PR #93, opened 2016-04-15
Case-insensitive treating of user headers

rouzwawi:
@pettermahlen @daenney

### DIFF
--- a/apollo-api/src/main/java/com/spotify/apollo/RequestValue.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/RequestValue.java
@@ -88,7 +88,9 @@ abstract class RequestValue implements Request {
   @Override
   public Request withHeaders(Map<String, String> additionalHeaders) {
     Map<String, String> headers = new LinkedHashMap<>(headers());
-    headers.putAll(additionalHeaders);
+    for (Map.Entry<String, String> header : additionalHeaders.entrySet()) {
+      headers.put(header.getKey().toLowerCase(), header.getValue());
+    }
     return create(method(), uri(), parameters(), headers, service(), payload(), ttl());
   }
 
--- a/apollo-api/src/test/java/com/spotify/apollo/RequestTest.java
+++ b/apollo-api/src/test/java/com/spotify/apollo/RequestTest.java
@@ -122,8 +122,8 @@ public class RequestTest {
 
     assertThat(requestWithHeader("/foo", "old", "value").withHeaders(newHeaders).headers(),
                is(ImmutableMap.of("old", "value",
-                                  "newHeader", "value1",
-                                  "newHeader2", "value2")));
+                                  "newheader", "value1",
+                                  "newheader2", "value2")));
   }
 
   @Test
@@ -132,7 +132,32 @@ public class RequestTest {
 
     assertThat(requestWithHeader("/foo", "old", "value").withHeaders(newHeaders).headers(),
                is(ImmutableMap.of("old", "value2",
-                                  "newHeader", "value1")));
+                                  "newheader", "value1")));
+  }
+
+  @Test
+  public void shouldTreatHeadersCaseInsensitivelyOnLookup() throws Exception {
+    Request request = Request.forUri("/foo")
+        .withHeader("User-Agent", "value");
+
+    assertThat(request.header("User-Agent"), is(Optional.of("value")));
+  }
+
+  @Test
+  public void shouldTreatHeadersCaseInsensitively() throws Exception {
+    Map<String, String> newHeaders = ImmutableMap.of(
+        "Accept", "application/json",
+        "raNdoM", "blabla");
+
+    Request request = Request.forUri("/foo")
+        .withHeader("User-Agent", "value")
+        .withHeaders(newHeaders);
+
+    assertThat(request.headers(),
+               is(ImmutableMap.of(
+                   "user-agent", "value",
+                   "accept", "application/json",
+                   "random", "blabla")));
   }
 
   @Test
--- a/apollo-api/src/main/java/com/spotify/apollo/Request.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/Request.java
@@ -72,9 +72,12 @@ public interface Request {
 
   /**
    * A header of the request message, or empty if no header with that name is found.
+   *
+   * Header names are case-insensitive. A lookup for {@code "User-Agent"} and {@code "user-agent"}
+   * will return the same header.
    */
   default Optional<String> header(String name) {
-    return Optional.ofNullable(headers().get(name));
+    return Optional.ofNullable(headers().get(name.toLowerCase()));
   }
 
   /**
@@ -111,6 +114,8 @@ public interface Request {
   /**
    * Creates a new {@link Request} based on this, but with an additional header.
    *
+   * Header names are case-insensitive.
+   *
    * @param name  Header name to add
    * @param value  Header value
    * @return A request with the added header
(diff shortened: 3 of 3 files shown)


pettermahlen:
Should also add a comment about the fact that header keys returned by the ```headers()``` method are always lower-case.

apollo-api/src/test/java/com/spotify/apollo/RequestTest.java
```diff
@@ -132,7 +132,32 @@ public void shouldReplaceExistingHeader() throws Exception {
...
+  @Test
+  public void shouldTreatHeadersCaseInsensitivelyOnLookup() throws Exception {
+    Request request = Request.forUri("/foo")
+        .withHeader("User-Agent", "value");
+
+    assertThat(request.header("User-Agent"), is(Optional.of("value")));
+  }
+
+  @Test
+  public void shouldTreatHeadersCaseInsensitively() throws Exception {
```
pettermahlen: Change name to 'shouldConvertHeaderNamesToLowerCase()' ?

daenney:
For reference in case this ever comes up. We need to tread headers case insensitively according to https://tools.ietf.org/html/rfc7230#section-3.2 and in HTTP/2 headers are all lower case by default.

apollo-api/src/test/java/com/spotify/apollo/RequestTest.java
rschildmeijer: Do we really want to lose this information?

apollo-api/src/main/java/com/spotify/apollo/RequestValue.java
```diff
@@ -88,7 +88,9 @@ public Request withHeader(String name, String value) {
   @Override
   public Request withHeaders(Map<String, String> additionalHeaders) {
     Map<String, String> headers = new LinkedHashMap<>(headers());
-    headers.putAll(additionalHeaders);
+    for (Map.Entry<String, String> header : additionalHeaders.entrySet()) {
+      headers.put(header.getKey().toLowerCase(), header.getValue());
+    }
```
udoprog: Is the right approach? This would effectively tie this class to only be usable with protocols where case-insensitivity is required for headers.

udoprog:
Cool this is happening. This allows me to get rid of hackery in my filters.

apollo-api/src/main/java/com/spotify/apollo/RequestValue.java
daenney: Can Apollo as it stands even handle protocols other than HTTP? It seems to be very much built up around HTTP and the lifecycle of an HTTP request.

daenney: To @udoprog's point though, when looking up the header all that would really be required is to do a `s1.equalsIgnoreCase('s2')` and you can keep storing it exactly in the format that came in, which could be useful for other purposes.

udoprog: I don't know the relevance. In #71 I suggest that service implementations provide their own implementations of `Request` to deal with this.

apollo-api/src/test/java/com/spotify/apollo/RequestTest.java
pettermahlen: I personally don't think header name case is relevant information. The reason it came up specifically is that HTTP/2 mandates lower-case headers: https://http2.github.io/http2-spec/#rfc.section.8.1.2. There's precedent also in how NodeJS treats header casing: https://github.com/nodejs/node-v0.x-archive/issues/1954 - I like the quote about how 'protocol artefacts are not information'.

pettermahlen:
I've done some thinking about the potential impact of this change. Here's the data I've gathered so far:

- The Apollo ```Request.headers()``` and ```Response.headers()``` methods return a ```Map<String, String>``` with undefined case for the keys/header names.
- Case-sensitive get/comparisons of the keys in the above map is broken, for HTTP/1.1 and HTTP/2 both.
- Apollo's API is broken since it returns a headers map without well-defined header name casing semantics. Netty's ```HttpRequest```, in contrast, returns a 'map-like' type called ```HttpHeaders```, which handles casing but is more complex.
- Other methods in ```Request``` and ```Response``` can be made case-safe without impact on callers.
- There may be a choice between taking a performance/complexity of implementation hit in Apollo (forever, more or less), or an upgrade/issues hit for users upgrading to, for instance, a version of Apollo where all headers are lower-case.
- Our internal survey shows:
  - 98.5% neutral/positive to lower-casing headers (1 respondent mildly opposed)
  - 94% of respondents don't want header name casing to be significant.
  - 97% are OK with losing the original case information.
- codesearch shows about 500 invocations of a method called ```.headers()``` in Spotify-internal code. Most of those are in tests, some of those are duplicates (from forks), and some are not Apollo requests. So the volume of internal invocations is probably manageable. I went through and roughly classified the files these hits occurred in and got the following results (number of files in each class):
  - Test code: 70 
  - Forks/experiments: 27
  - Not Apollo ```.headers()```: 26
  - Apollo, but not problematic (due to not using the resulting maps in a case-sensitive way): 20
  - Risky: 23
- If that classification is roughly correct, this change would break a bunch of tests, and would require special attention in about 23 places. (Broken tests are pretty much OK, because they're easy to fix and it will be obvious that they're broken. We don't want to risk silently changing production code behaviour.)

I can see three main approaches:

1. Make the 'always-lower-case' change currently in this PR and pay special attention to upgrading/fixing the places where the unsafe ```headers()``` method is invoked.
   - Keeps the API clean and simple.
   - Leads to a risk of not upgrading all internal callers (and of course, we won't be upgrading any open source users).
   - Makes the upgrade process a bit painful for everyone who has case-sensitive tests.
1. Deprecate, but don't change the behaviour of the ```headers()``` method. Instead, add a separate method that is safe. 
   - Increases internal code complexity.
   - Messes up the API.
   - Leads to some performance hit, if nothing else then due to increased allocation.
1. A major version upgrade, changing the signature of the ```headers()``` method (probably to returning something similar to the Netty ```HttpHeaders``` class).
   - Keeps the API simple.
   - Is a pretty small change in a new major version, and I'm not sure if there are other things we know we want to include in a new major version. But there's plenty of numbers out there, so that shouldn't be a big concern.
   - makes the upgrade process a bit painful for everyone who relies on getting a ```Map<String, String>``` from the headers() method, but at least errors will be found by the compiler rather than at runtime.
   - Adds internal implementation complexity.

I don't have a strong preference as to which way to go, yet. @ccarpita, do you have any input or opinions?

daenney:
I'm not a huge fan of 2. As you said it'll add complexity but will still require people to manually go spelunking through their code and find all the places this is happening and switch over to the new method of dealing with headers. When we get to the next major, do we then clean the API up and need everyone to go back?

At least with option 1 we'll have some automated "discovery" of that problem thanks to the tests that'll fail though it's less safe than option 3.

mattnworb:
With option 1 I would worry about the places where problems caused by change in behavior/casing are not caught by tests because the tests don't exist. 

2 and 3 seem like the best choices to me. I like 3 better as a compiler error caused by signature change is a lot more obvious than a compiler warning about deprecated methods.

ccarpita:
I have to agree with the breaking change approach, given the severity of issues which may appear nondeterministically at runtime.  If the major is bumped, along with release notes explaining the change, this provides a local optimum to the user, who can understand the reasoning and patch their dependent code accordingly.

rouzwawi:
I'm for 1 or 3. If we do 3, I would like to clean up any existing deprecated classes and other adjustments we know we want to do.

pettermahlen:
Moving forward with option 3. For a list of issues we're planning on including in apollo 2.0, see https://github.com/spotify/apollo/milestones/2.0.0 - feel free to create others if there are changes you'd like to see.

rouzwawi:
I will close this PR in favour of #96

PR #94, opened 2016-04-22
allow for basic configuration of HttpClient

mattnworb:
Allow the underlying OkHttpClient instance to be configured with a few
basic properties from the Config, if specified. The configurable
properties are:

- connect timeout - `http.client.connectTimeout`
- read timeout - `http.client.readTimeout`
- write timeout - `http.client.writeTimeout`
- connection pool: max idle connections - `http.client.maxIdleConnections`
- connection pool: keep alive duration - `http.client.keepAliveDuration`
- max async requests - `http.client.async.maxRequests`
- max async requests per host - `http.client.async.maxRequestsPerHost`

The default value from OkHttpClient is used for any unspecified
properties (which is the behavior previous to this commit).

Also renamed the factory method `HttpClient.create()` to
`HttpClient.createDefault()` to make it clear that the instance being
returned is using all of the default configuration properties.


mattnworb:
I am not that familiar with Guice so I would appreciate any review of how I am using it here. 

For instance, it feels weird to me that I had to create a HttpClientProvider when all it does is wire the OkHttpClient instance into the HttpClient constructor. Is there a way to avoid declaring a class for just that? I originally had all of the OkHttpClient construction in the HttpClientProvider class, but I wanted to be able to add a test that the Config keys were properly parsed into custom values of things like `OkHttpClient.getConnectTimeout()` but couldn't see an easier way to do that.

Also, is it problematic to have something like `bind(OkHttpClient.class).to(OkHttpClientProvider.class)` when the former is a 3rd party class that someone could conceivably already be binding in their own service?

mattnworb:
btw, the license-maven-plugin seems to enforce that the license header has trailing spaces in lines that only contain `*`, which some editors might automatically remove 😕

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClientDecoratorProvider.java
```diff
@@ -21,12 +21,21 @@
...
 import javax.inject.Provider;
 
 class HttpClientDecoratorProvider implements Provider<ClientDecorator> {
 
+  private final HttpClient httpClient;
+
+  @Inject
+  HttpClientDecoratorProvider(final HttpClient httpClient) {
+    this.httpClient = httpClient;
+  }
```
mattnworb: this also feels weird for the same reason as HttpClientProvider

pettermahlen:
I'm also not a Guice expert, but LGTM. @rouzwawi ?

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClientDecorator.java
```diff
@@ -24,17 +24,14 @@
...
 
-  private static final HttpClientDecorator INSTANCE = new HttpClientDecorator();
+  private final HttpClient httpClient;
 
-  HttpClientDecorator() {
-  }
-
-  public static HttpClientDecorator create() {
-    return INSTANCE;
+  HttpClientDecorator(HttpClient httpClient) {
```
rouzwawi: `@Inject` on this and you can get rid of the providers, I think.

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClientProvider.java
```diff
@@ -0,0 +1,39 @@
...
+  private final OkHttpClient okHttpClient;
+
+  @Inject
+  HttpClientProvider(OkHttpClient okHttpClient) {
+    this.okHttpClient = okHttpClient;
+  }
+
+  @Override
+  public HttpClient get() {
+    return new HttpClient(okHttpClient);
```
rouzwawi: This constructor already has an `@Inject` annotation, so it should suffice to just bind the class in the module, as `bind(HttpClient.class)`.

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClientModule.java
```diff
@@ -38,6 +39,9 @@ public static ApolloModule create() {
   protected void configure() {
     Multibinder.newSetBinder(binder(), ClientDecorator.class)
         .addBinding().toProvider(HttpClientDecoratorProvider.class);
+
+    bind(HttpClient.class).toProvider(HttpClientProvider.class);
+    bind(OkHttpClient.class).toProvider(OkHttpClientProvider.class);
```
rouzwawi: Binding these without the `.to...` part should make Guice find injectable constructors.

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
```diff
@@ -38,11 +40,18 @@
...
     this.client = client;
   }
 
+  @Deprecated
+  /** @deprecated use {@link #createDefault()} instead. */
   public static HttpClient create() {
+    return createDefault();
+  }
+
+  public static HttpClient createDefault() {
```
rouzwawi: Why did this have to change name?

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
```diff
@@ -20,6 +20,8 @@
 package com.spotify.apollo.http.client;
 
 import com.spotify.apollo.environment.IncomingRequestAwareClient;
+
+import com.google.inject.Inject;
```
rouzwawi: prefer using `javax.inject.Inject`

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClientProvider.java
rouzwawi: As done here: https://github.com/spotify/apollo/blob/master/modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/HttpServerModule.java#L51

pettermahlen:
Another thing: it would be great to add documentation of the configuration settings in a README for the client. There's no such readme at the moment.

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
mattnworb: It didn't have to, but I wanted to make it clear that it was creating a HttpClient that was unconfigured / with all of the default configuration values

mattnworb: and now that I say that I am not sure why I didn't just call it `createUnconfigured()` 😄

mattnworb:
wow coveralls is precise

> ❌  Coverage decreased (-0.05%) to 67.542%

mattnworb:
@rouzwawi @pettermahlen thanks for the review, I added new commits which simplified things and removed the need for one of the Providers, and also added a README for the module.

pettermahlen:
👍 from me

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClientDecoratorProvider.java
rouzwawi: Can't this class be dropped all together now that HttpClientDecorator has an injectable constructor?

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
rouzwawi: Ah, makes sense, and `createUnconfigured()` makes more sense :)

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClientDecoratorProvider.java
mattnworb: Looks like you are correct - see the newest commit 9911529

PR #96, opened 2016-04-26
Make request and response header names lower-case

pettermahlen:
This change introduces a Headers type so as to break backwards
compatibility and trigger compiler errors if the Request.headers()
or Response.headers() methods are called expecting a Map as the
return value.

Header names are lower-cased, to simplify case comparisons and be
compatible with HTTP/2. A part of the reasoning around making
this simplistic change is that lower-casing is very cheap for
a string that is already lower-case, and thus doing the lower-
casing always will only lead to a performance hit the first time
a header is processed. Headers that are propagated internally
will always be lower case when handled by downstream services.

This is an alternative solution to #93.


pettermahlen:
@rouzwawi @ccarpita @mattnworb @zalenski opinions?

apollo-api/src/main/java/com/spotify/apollo/HeadersValue.java
```diff
@@ -0,0 +1,77 @@
...
+  public Optional<String> get(@Nonnull String name) {
+    return Optional.ofNullable(asMap().get(name.toLowerCase()));
+  }
+
+  static Headers create(Map<String, String> headers) {
+    Map<String, String> allLowerCaseKeys = headers.keySet().stream()
+        .filter(HeadersValue::hasNonLowerCase)
+        .findAny()
+        .map(lowerCaseKeys(headers))
+        .orElse(headers);
```
mattnworb: this stream operation seems confusing to me:

1. it would seem to filter out any keys that are all lower-case 
2. and the map operation ignores the input (the `s` in `s ->`)

is the goal to avoid modifying the map in the happy case when the keys are already all lower case?

mattnworb: I mean, I see in the tests that it works, it was just confusing as to how it works until I ran the tests manually to see 😄

mattnworb:
Overall seems sensible to me, although I don't have much of a dog in this fight 👍

pettermahlen:
The failed build appears to be due to a problem with codecov.io. They seem to be having problems at the moment, according to: http://status.codecov.io/, which says there are intermittent failures due to a systems upgrade since a couple of days.

pettermahlen:
@mattnworb thanks for the feedback, i've tried to clarify the implementation a bit - better?

ccarpita:
I did happen to have a dog in this fight, and am pleased with the results :)

FWIW the updated lowercase-checking implementation seemed clear to me at first read.

mattnworb:
@pettermahlen 👍  on improved clarity!

PR #102, opened 2016-05-17
Fix typo in code example

nresare:
Provide cassandraClusterConnection as argument to the AddressResource constructor
instead of nonexistent responseText variable


rouzwawi:
Thanks!

PR #104, opened 2016-05-18
Update example to match updated Response

nresare:
Replace Response.statusCode(), which seems to have vanished, with Response.status() in README.md example


pettermahlen:
Replaced by #106

PR #109, opened 2016-05-21
Remove instance executors

pettermahlen:
The idea is that executors are easy to provide at the application level, if needed, and that the ones provided by Apollo are very rarely used in practice. So it's a feature of Apollo that doesn't carry its weight.

This PR also does some minor cleanup of the ServiceImpl class.

### DIFF
--- a/apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
+++ b/apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
@@ -21,7 +21,6 @@ package com.spotify.apollo.core;
 
 import com.google.common.base.Function;
 import com.google.common.base.MoreObjects;
-import com.google.common.collect.FluentIterable;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Ordering;
@@ -56,9 +55,9 @@ import java.util.Objects;
 import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
-import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Collectors;
 
 import joptsimple.BuiltinHelpFormatter;
 import joptsimple.OptionException;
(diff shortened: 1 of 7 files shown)


pettermahlen:
@rouzwawi @zalenski

apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
```diff
@@ -515,33 +479,7 @@ public Service build() {
     }
   }
 
-  static class ExecutorServiceCloseable implements Closeable {
```
rouzwawi: I've had to write code similar to this many times in services where I create my own executors. It would be nice if there was an easy was to add application level executors to the Apollo shutdown cycle, instead of writing this code each time.

pettermahlen: Specifically for closing executors, you mean? If so, would something like
```
  environment.closer().register(ExecutorCloser.of(executor));
```
work?

rouzwawi: Yep that would be perfect

pettermahlen:
Turns out closing an executor service is somewhat complicated, since you're not allowed to throw InterruptedExceptions from a close() method. But here's some code that does that.

mattnworb:
👍

apollo-core/src/test/java/com/spotify/apollo/core/ServiceImplTest.java
```diff
@@ -366,16 +375,8 @@ public void testSignalShutdown() throws Exception {
...
 
-      // Force predictable concurrency of signal and main loop start
-      final CountDownLatch countDownLatch = new CountDownLatch(1);
-      instance.getExecutorService().submit(() -> {
-        countDownLatch.await();
-        Thread.sleep(200);
-        instance.getSignaller().signalShutdown();
-        return null;
-      });
+      instance.getSignaller().signalShutdown();
```
dflemstr: I think the point of this test was to `.signalShutdown()` while the main thread blocks on `.waitForShutdown()`. You'll probably have to replace the executor by another executor to get the same behavior.

apollo-extra/pom.xml
```diff
@@ -67,6 +67,11 @@
             <artifactId>mockito-all</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.hamcrest</groupId>
+            <artifactId>hamcrest-library</artifactId>
```
dflemstr: Would encourage using Hamcrest 2 for new code

apollo-core/src/test/java/com/spotify/apollo/core/ServiceImplTest.java
pettermahlen: Yes. I didn't see much added value in doing shutdown and wait on separate threads, though, so I simplified the test case. Do you think there is value there?

pettermahlen:
@rouzwawi any further comments before I merge?

rouzwawi:
LGTM :+1:

PR #110, opened 2016-05-21
Add new StubClient methods for defining request mappings

### LINKED ISSUE #62: StubClient respondTo order
Copied from internal repo, by jwiklund:

The stub client uses the first stubbed response, it would be convenient if it took the last match to allow specific tests to override generic stubs.

eg:

@Before
stubClient.respond(OK).toEverything()

@Test
stubClient.respond(FAIL).toSomethingSpecific()

pettermahlen:
Fixes #62

The rationale for adding new methods and deprecating the old, while retaining their behaviour is that changing the behaviour seems to potentially cause unnecessary upgrading pain.

The rationale for not including the ability to specify a duration for delaying the response in the new methods is that it's not a very common use case, and specifying the delay is easy through the ```Responses``` class.

### DIFF
--- a/apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
+++ b/apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
@@ -259,7 +259,11 @@ public class StubClient implements Client, Closeable {
 
   /**
    * Immutable response builder.
+   *
+   * @deprecated in favour of using the {@link #addMapping(Matcher, Response)} or
+   * {@link #addMapping(Matcher, ResponseSource)} methods.
    */
+  @Deprecated
   public class StubbedResponseBuilder {
     @Nullable
     private final ResponseWithDelay responseWithDelay;
--- a/apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
+++ b/apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
@@ -38,7 +38,6 @@ import java.io.Closeable;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
-import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionStage;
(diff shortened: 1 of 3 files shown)


jwiklund:
:+1:

pettermahlen:
@jo-ri WDYT?

apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
```diff
@@ -275,13 +319,48 @@ private ResponseSource responseSource() {
 
   }
 
+
+  public class WhenResponseBuilder {
```
jo-ri: As objects of this class should not be reused, I think supplying both arguments to addMatcherFirst at the same time is a little bit clearer.
If this class will remain I think it could have a different name as 'builder' may be misleading considering that it modifies the StubClient.

apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
```diff
@@ -275,13 +319,48 @@ private ResponseSource responseSource() {
...
+     * Respond with responses and delays supplied by the provided ResponseSource to any matching
+     * request.
+     */
+    public void respond(ResponseSource responseSource) {
+      addMatcherFirst(requestMatcher, responseSource);
+    }
+  }
+
   @AutoValue
-  public static abstract class RequestResponsePair {
```
jo-ri: Is API breaking change ok?

apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
```diff
@@ -275,13 +319,48 @@ private ResponseSource responseSource() {
...
     public abstract Request request();
     public abstract Response<ByteString> response();
 
     public static RequestResponsePair create(Request request, Response<ByteString> response) {
       return new AutoValue_StubClient_RequestResponsePair(request, response);
     }
   }
+
+  @AutoValue
+  static abstract class MatcherResponseSourcePair {
```
jo-ri: I think a private static final class with final members would cause the least cognitive overhead in this case.

apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
```diff
@@ -193,7 +204,12 @@ public void clear() {
    * Configure a constant (i.e., all matching requests will always result in the same response)
    * response for some request. The returned builder allows configuration of payload (default
    * no payload) and delay before the response is sent (default 0).
+   *
+   * When using this method to define request to response mappings, the mappings will be evaluated
+   * in the order which they were added.
+   * @deprecated in favour of {@link #when(Matcher)} or {@link #when(String)}.
    */
+  @Deprecated
```
jo-ri: What is the reason the respond methods are deprecated?

apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
pettermahlen: Nope, that was a mistake!

pettermahlen: Well, to be precise - I'm OK with breaking the API, since this is something that goes into 2.0, but this class still needs to be public, for people to be able to see what the StubClient has done.

apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
pettermahlen: Done

apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
pettermahlen: It's because I don't like solutions that allow you to do the same thing in two different ways. I prefer being opinionated about how things should work, it makes things simpler. So I would like to recommend people to use only the new version of the API.

apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
pettermahlen: Done - how does the new version look?

apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
jo-ri: Ok, great! I think we should also deprecate StubbedResponseBuilder and remove it along with the respond methods.

PR #111, opened 2016-05-23
Add Codenvy factory to test this project including example

benoitf:
### Codenvy Workspace Developer

Hello,

this pull request introduces a new **_“Codenvy workspace developer”**_ button in the README of the project.

[Contribute](http://beta.codenvy.com/f?id=tp3jw04xnwzhrz92)

This button enables anyone to test or contribute to the project without installing anything.

When a developer presses the button, a new workspace is created:
- A project runtime is created in a Docker container running in the Codenvy cloud (for free). You can customize this container using a Dockerfile.
- Source code is automatically cloned into the workspace as a “local” copy.
- A command set is added so users can build / run / debug the project with a single click.
- The project is automatically run so a new user can experience it hassle-free.
- A pull request panel in the IDE simplifies the process of creating a PR for new users. Experienced users can use the git menu in the IDE, or issue git commands from the built-in terminal.

This does not require **_any changes**_ to the source code and has **_no cost**_ to you or your users.

Here is a youtube video link to check result:

[Youtube link](http://www.youtube.com/watch?v=Oh5cWM5zYP0)

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,8 @@ Apollo
 [![Codecov](https://img.shields.io/codecov/c/github/spotify/apollo.svg)](https://codecov.io/gh/spotify/apollo)
 [![Maven Central](https://img.shields.io/maven-central/v/com.spotify/apollo-parent.svg)](https://search.maven.org/#search%7Cga%7C1%7Cg%3A%22com.spotify%22%20apollo*)
 [![License](https://img.shields.io/github/license/spotify/apollo.svg)](LICENSE.txt)
+[![Contribute](http://beta.codenvy.com/factory/resources/codenvy-contribute.svg)](http://beta.codenvy.com/f?id=tp3jw04xnwzhrz92)
+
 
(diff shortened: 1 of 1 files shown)


pettermahlen:
I'm not completely sure I understand the use case for this? It's in a way never the case that something has 'no cost', since there's clutter. :) That means I think we should primarily merge in things that we're going to be using.

Also, without the button, what are the other ways that users can use Codenvy, if any?

TylerJewell:
@pettermahlen - In this case, it would be a service that you can provide to unknown future contributors who would like to make pull requests or other suggestions to this repository. The button enables those users to get a workspace that is pre-configured to build and run the project. Also, they can commit back directly if they have commit rights, or issue a pull request (where the system will auto-fork, auto-branch for the user, if necessary).  A side affect of this behavior is that for those that are just users of this technology, they could have an alternative way to run the software.

Codenvy offers free workspaces for open source projects for their committers / contributors. There are no strings attached.

These workspace configurations are being used through various Java projects - tomee, vert.x, eclipse. This is based upon Eclipse Che, a next-generation Eclipse IDE that is cloud-based.

martina-if:
We considered this, but since we probably won't use it we decide it to close this pull request.

PR #116, opened 2016-06-14
DRAFT: Typed configuration

pettermahlen:
This PR is NOT to be merged, it's for discussing an early proposal for replacing the current, string-based configuration solution with a typesafe and navigable/discoverable one. 

The overall purpose is to make it easier for Apollo users to configure their Apollo service, through:
1. Providing a schema/definitive set of all Apollo-defined configuration keys.
2. Making it easier to understand the impact of changing the value of an Apollo configuration key by making it easier to find where the configuration key is used.
3. Reducing the risk of making incorrect changes through, for instance, typos in a config file.
4. Making it easier to understand why a configuration key has a particular value; how that value was derived.

The proposal is to treat service-specific configuration settings separately, since we make the assumption that Apollo users are capable of understanding their own configurations more easily than the framework configuration settings. So configuration keys like `http.server.workerThreads` and `apollo.logIncomingRequests` are in scope for the bulk of this document, but `myservice.psql.name` is not. 

Highlights of the current changes:
- Define configuration types for core Apollo modules (`ApolloConfig`, `JettyHttpServerConfiguration`, etc.)
- For each assembly, define an overall configuration type (`HttpServiceConfiguration`).
- Load service-specific configuration using vanilla Typesafe config ([ConfigFactory.load()](https://github.com/typesafehub/config#standard-behavior)), with no Apollo-specific override mechanisms.
- Permit overriding of Apollo-specific configuration:
  -  programmatically (TBD), 
  -  via system properties (TBD),
  - and via environment variables (TBD)
- Remove all or most Apollo command line arguments


martina-if:
Hi @pettermahlen, this was an interesting PR, will you have time to finish it and want us to keep it open or should we close it until we do more work on apollo 2.0?

pettermahlen:
I won't be able to finish it, I'm afraid. Configuration was one of the most asked question types, and I think this would help make it more accessible/discoverable, so it would be a good thing to do. Whether you prefer to keep the PR open or close it for now is up to you - I will leave the branch untouched in case you want to start working on it again.

martina-if:
Thanks @petter. I added a link to this PR to the apollo roadmap document and I will close this for now.

PR #117, opened 2016-06-14
Add Exploratory annotation

pettermahlen:
With the intention of being able to flag parts of the API as less stable than the rest, for user convenience.


pettermahlen:
I'm not completely sure how beneficial this is. I'm thinking it can be useful to mark APIs that we're still exploring how to do best - such the the Entity API, and such as a future Websockets API - so that users can understand that it's likely to be changed.

pettermahlen:
Merging this; we'll find out in a year or two if this annotation is useful. If not, it's at least extremely unlikely to be harmful. :)

PR #122, opened 2016-06-29
Add semantic metrics module

pettermahlen:
This is the first step in a series of changes to add metrics to
HttpService assemblies. Includes:

- The to-be-deprecated https://github.com/spotify/apollo-metrics project added as a module.
- Using the metrics module in `HttpService`
- Some name changes and other cleanup of apollo-metrics


pettermahlen:
@rouzwawi @jo-ri

modules/metrics/src/main/java/com/spotify/apollo/metrics/InstrumentingClient.java
```diff
@@ -0,0 +1,50 @@
...
+import com.spotify.apollo.request.TrackedOngoingRequest;
+
+import java.util.concurrent.CompletionStage;
+
+import okio.ByteString;
+
+/**
+ * A {@link Client} that collects metrics on outgoing calls
+ */
+class InstrumentingClient implements Client {
```
jo-ri: Should we instrument also the client provided in the environment (using ClientDecorator)?

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
```diff
@@ -0,0 +1,118 @@
...
+          @Override
+          protected Ratio getRatio() {
+            return ratioSupplier.get();
+          }
+        });
+  }
+
+  @Override
+  public void fanout(int requests) {
+    metricRegistry.histogram(fanoutId).update(requests);
```
jo-ri: Fetch once and keep instead of id

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
```diff
@@ -0,0 +1,118 @@
...
+
+    StatusType.Family family = status.family();
+    if (family != INFORMATIONAL && family != SUCCESSFUL) {
+      sentErrors.mark();
+    }
+  }
+
+  @Override
+  public TimerContext timeRequest() {
+    return new SemanticTimerContext(metricRegistry.timer(timeRequestId).time());
```
jo-ri: Fetch once and keep instead of id

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  @Override
+  public void fanout(int requests) {
+    metricRegistry.histogram(fanoutId).update(requests);
+  }
+
+  @Override
+  public void responseStatus(StatusType status) {
+    metricRegistry.meter(countRequestId.tagged("status-code", String.valueOf(status.code()))).mark();
```
jo-ri: Could consider keeping metrics in an array (MetricId javadoc comments this is slow), but it could very well be considered premature optimisation.

jo-ri:
+1 (in the future we should consider making this library easier to use outside guice, and combine required metric tags (e.g. stat) into a single 'what' )

modules/metrics/src/main/java/com/spotify/apollo/metrics/InstrumentingClient.java
pettermahlen: Yes, we should, in another PR, I think.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
pettermahlen: This is changed in the next PR anyway, so leaving as-is for now to reduce the amount of merge conflicts. :)

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
pettermahlen: Fixed.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
udoprog: This has another bad side-effect. The metric will not exist until at least one of the specific error code has been reported. It makes for a very poor user experience.

My suggestion is to pre-define the meters for each significant error category/code and use them, like: `2xx`, `3xx`, `4xx`, `404`, `5xx`, `503`. This way, you avoid the first-report creation problem.

modules/metrics/README.md
```diff
@@ -0,0 +1,175 @@
...
+   * Middleware to track response payload size in a Histogram,
+   * tagged with an endpoint tag set to the given endpoint name.
+   */
+  public Middleware<AsyncHandler<Response<ByteString>>, AsyncHandler<Response<ByteString>>>
+      responsePayloadSizeHistogram(String endpointName) {
+
+    final MetricId histogramId = MetricId.build()
+        .tagged("service", serviceName)
+        .tagged("endpoint", endpointName)
+        .tagged("what", "endpoint-response-size");
```
udoprog: This metric name could easily be misconstrued as the total response size. Maybe call it `endpoint-response-payload-size`?

pettermahlen: Good point - the doc is outdated, but I changed the names in https://github.com/spotify/apollo/pull/123/commits/3b30b9ce482580ce23ee17959055d39b099b57a1. Make sense?

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
pettermahlen: Good point - what do you think about https://github.com/pettermahlen/apollo/commit/ff1cf0bfabcfe9173cb0d548e045b6bdb714acb4 ?

pettermahlen: BTW; avoiding status-family, which we used internally because there are too many status codes, in favour of a configurable set of status codes to precreate. In general, for a given service, you will know which codes you're generating and it's often the case that you want the resolution individual codes give you.

udoprog: I'd prefer an implementation that completely disallows the dynamic creation of metric descriptors. One could allow for a set of detailed status code through configuration or convention, but still mark the family-style status codes (e.g. `2xx`). Any status code that is  outside of configuration would not be created.

Dynamically created metric descriptors has caused a lot of end-user pain in the past. While the suggested approach improves it, I'd prefer excluding them completely.

pettermahlen: Created #126 to track this; we need to discuss in a better place.

PR #127, opened 2016-07-01
bump version to 1.2.0

jo-ri:


jo-ri:
@rouzwawi

rouzwawi:
I don't know if this is exhaustive, but it you say so then 👍

PR #128, opened 2016-07-05
Add status-family tag to metric tagged with what=endpoint-request-rate

jo-ri:


jo-ri:
Think we could close this until it's requested

PR #130, opened 2016-07-11
set provided scope for auto-value dependency

### LINKED ISSUE #121: apollo-bom doesn't use scopes in dependencyManagement
I noticed in a project I am working on that a dependency on `apollo-test:1.1.0:test` ends up dragging in `com.google.auto.value:auto-value:1.1`. I expected the latter to have `<scope>provided</scope>` in apollo-test, since AFAIK code generated by auto-value has no dependencies on auto-value (and scope=provided is what [the auto-value docs recommend](https://github.com/google/auto/blob/master/value/userguide/index.md).

But I noticed that in the apollo-bom, [none of the modules under the comment for "provided scope" actually have scope=provided](https://github.com/spotify/apollo/blob/v1.1.0/apollo-bom/pom.xml#L223-L238). Is that an oversight? Or a limitation of the bom concept?

mattnworb:
This prevents users of apollo-test from having auto-value pulled into
their project as a compile-scope dependency.

Fixes #121.

### DIFF
--- a/apollo-test/pom.xml
+++ b/apollo-test/pom.xml
@@ -44,6 +44,7 @@
         <dependency>
             <groupId>com.google.auto.value</groupId>
             <artifactId>auto-value</artifactId>
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.hamcrest</groupId>


mattnworb:
@rouzwawi

PR #131, opened 2016-07-15
Have okhttp-client propagate headers from from apollo Request

nresare:
Headers from the Apollo Request object would previously not propagate to the remote endpoint when doing http requests utilising the okhttp-client module. This change, with test case, makes sure that the headers propagate from the Apollo Request object to the okhttp equivalent.


nresare:
ping @dflemstr

nresare:
Superseded by https://github.com/spotify/apollo/pull/132

PR #132, opened 2016-07-15
Okhttp client auth request propagation

nresare:
1. Since this modifies the functionality in https://github.com/spotify/apollo/pull/131 I created this pull request on top of that commit.
2. At some point I guess we might want to benchmark this code to figure out if it would make sense to trade branching for short lived object creation, as it gets run for every outgoing request.


rouzwawi:
I'm surprised this was not already in place. Thanks for contributing it.

PR #133, opened 2016-07-18
Add Route#withDocString(DocString)

odsod:
Better API when loading documentation from bundled resources, or
partially generating doc strings from e.g. Protobuf payload descriptors.


rouzwawi:
:+1:

PR #135, opened 2016-07-22
Dont use executor in StubClient if delay <= 0

spkrka:
It is useful for tests to have few threads to be more
deterministic. There is no reason to schedule responses
on an executor if the delay <= 0


rouzwawi:
👍

PR #137, opened 2016-07-22
Close down okhttp dispatcher executor

rouzwawi:
This seems to be the way one should close the okhttp client according to: https://github.com/square/okhttp/blob/master/okhttp/src/main/java/okhttp3/OkHttpClient.java#L90-L100

I've tested this with a CLI that earlier would hang the JVM for about 30 seconds before actually exiting. Looking at the threads, I saw that one non-deamon thread called `Dispatcher` was not stopped.

### DIFF
--- a/modules/jetty-http-server/pom.xml
+++ b/modules/jetty-http-server/pom.xml
@@ -63,7 +63,6 @@
         <dependency>
             <groupId>com.squareup.okhttp</groupId>
             <artifactId>okhttp</artifactId>
-            <version>2.5.0</version>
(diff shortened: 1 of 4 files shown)


spkrka:
:+1:

mattnworb:
this will be very helpful, thanks @spkrka

spkrka:
@mattnworb all credit goes to @rouzwawi , I just reviewed it. :)

mattnworb:
d'oh I meant to comment on #135 !

PR #139, opened 2016-07-27
Populate Request#service from X-Calling-Service HTTP header

odsod:
Please help me find an appropriate place to document this.

### DIFF
--- a/modules/jetty-http-server/README.md
+++ b/modules/jetty-http-server/README.md
@@ -17,6 +17,12 @@ key | type | required | note
 
 _todo_ implement all config keys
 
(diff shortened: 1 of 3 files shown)


pettermahlen:
LGTM, except that I think the tests should validate the behaviour when the header is missing (Optional.empty() - not an error, which might happen if somebody is careless). Might also want to consider what should happen if the header value is the empty string. I guess that should be the same as if the header wasn't there? 

The best place to document this is probably in the jetty-http-server README file.

pettermahlen:
👍

PR #140, opened 2016-08-01
Explicitly state request rate in docs

davidxia:
for both endpoint-request-rate and dropped-request-rate.

### DIFF
--- a/modules/metrics/README.md
+++ b/modules/metrics/README.md
@@ -25,7 +25,7 @@ A Meter metric, per status-code, tagged with:
 |-------------|----------------------------|------------------------------------------------------|
 | what        | "endpoint-request-rate"    | Enable/disable with ENDPOINT_REQUEST_RATE            |
 | status-code | *                          | "200", "404", "418", etc.                            |
-| unit        | "request"                  |                                                      |
+| unit        | "request/s"                | Requests per second.                                 |
 
 ### Endpoint request duration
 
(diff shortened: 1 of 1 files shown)


modules/metrics/README.md
```diff
@@ -25,7 +25,7 @@ A Meter metric, per status-code, tagged with:
 |-------------|----------------------------|------------------------------------------------------|
 | what        | "endpoint-request-rate"    | Enable/disable with ENDPOINT_REQUEST_RATE            |
 | status-code | *                          | "200", "404", "418", etc.                            |
-| unit        | "request"                  |                                                      |
+| unit        | "request/s"                | Requests per second.                                 |
```
davidxia: I noticed that internally for us this is "message/s". Am I correct in assuming the open-source module without Spotify-flavoring will have "request/s" as its unit?

davidxia:
@pettermahlen

modules/metrics/README.md
pettermahlen: The `endpoint-request-rate` metric generated by internal Apollo in fact has the unit 'request', exactly like here. I think 'request/s' is a better unit, and slightly better than 'message/s'?

pettermahlen:
I think when making this change, we should also update the units used in the actual metrics tags as per the `SemanticServiceMetrics` class. Would you be able to add that? There seems to be other rates that have incorrect units (like the dropped requests rate, for instance).

jo-ri:
"/s" is appended to the 'unit' tags by ´FastForwardReporter´ so they should be defined without it in ´ SemanticServiceMetrics´.

davidxia:
@jo-ri Why does FastForwardReporter handle the appending of the "/s"? Why not just have `SemanticServiceMetrics` do that?

jo-ri:
@davidxia I don't know, I haven't worked with FastForwardReporter. Maybe @udoprog can shed some light?

udoprog:
Hey,

When some things are registered the system actually exposes multiple different things, sometimes with different units.
The most notable example is timer, which exposes histograms and meters. To make things consistent we made it so that the MetricId would only contain the singular unit and we derive the unit for the rates as described.

I'm open for discussing if this is the 'right' thing or not.

davidxia:
Hm, I don't know enough to have an opinion. As a user, I just wanted the docs to be more specific. Does this PR make sense anymore or should we close as the changes seem to be incorrect?

pettermahlen:
It does make sense, and that my suggestion of also updating the code was wrong. :) Merging.

PR #142, opened 2016-08-02
ensure logback-classic is an HttpService dependency

pettermahlen:
Also:
 - update documentation to reflect this
 - upgrade to the latest version of logback-classic
 - do some minor cleanup


pettermahlen:
@jo-ri WDYT?

jo-ri:
Perfect! +1

PR #147, opened 2016-08-05
Fix Content-Type utf-8 header

skoushan:
Should say "utf-8" instead of "UTF8" based on [W3C](https://www.w3.org/International/articles/http-charset/index)


pettermahlen:
Thanks!

PR #151, opened 2016-08-10
don't handle ERROR_DISPATCH requests from Jetty

pettermahlen:
If, for instance, an asynchronous request times out (configured
via org.eclipse.jetty.server.HttpChannelState.DEFAULT_TIMEOUT),
Jetty will eventually retry the request. We didn't handle that
case. This change means we check if Jetty has told us there was
a problem with the request, and if so, use that information to
return an error response

### DIFF
--- a/modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java
+++ b/modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java
@@ -44,6 +44,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import javax.servlet.AsyncContext;
+import javax.servlet.RequestDispatcher;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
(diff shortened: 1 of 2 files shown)


pettermahlen:
@jo-ri

sisidra:
Is it worth adding some documentation about this jetty timeout? Or maybe it is expected from users to know about such timeout and how to configure it?

modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java
```diff
@@ -74,6 +77,17 @@ public void handle(
...
 
+    if (baseRequest.getAttribute(ERROR_STATUS_CODE) != null) {
+      LOGGER.debug("Ignoring error dispatch request");
+      int statusCode = Integer.valueOf(baseRequest.getAttribute(ERROR_STATUS_CODE).toString());
+      Object message = baseRequest.getAttribute(ERROR_MESSAGE);
+      resp.sendError(statusCode, message == null ? null : message.toString());
+      return;
+    }
+
+    LOGGER.info("handling request {} with status code {}", baseRequest,
```
jo-ri: Last argument will always be null

modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java
```diff
@@ -74,6 +77,17 @@ public void handle(
       HttpServletRequest req,
       HttpServletResponse resp) throws IOException, ServletException {
 
+    if (baseRequest.getAttribute(ERROR_STATUS_CODE) != null) {
```
jo-ri: Maybe check baseRequest.getDispatcherType() == DispatcherType.ERROR instead?

jo-ri:
+1

pettermahlen:
Will solve this differently, having now done some homework on how `AsyncContext` works.

PR #155, opened 2016-08-15
allow ServiceHelper to resolve provided instances

udoprog:
This needed when writing integration tests to gain access to data access layers and such without code duplication.

For example.

``` java
final ServiceHelper helper = ServiceHelper.create(new Application(), "application")
        .withModule(new ApplicationModule())
        .withModule(new FakeDatabaseLayer());

helper.start();

final Database database = helper.resolve(Database.class);
```

### DIFF
--- a/apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelper.java
+++ b/apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelper.java
@@ -438,6 +438,23 @@ public class ServiceHelper implements TestRule, Closeable {
     return request(Request.forUri(uriString, method).withPayload(payload));
   }
 
+  /**
+   * Get an instance provided by one of the Apollo modules.
+   *
+   * @param type The type of instance to get
+   * @return     An instance of the given type.
+   * @throws java.lang.IllegalStateException if the service instance is not available.
+   */
+  public <T> T resolve(Class<T> type) {
+    final Service.Instance instance = this.instance;
+
+    if (instance == null) {
+      throw new IllegalStateException("service instance not available");
+    }
+
+    return instance.resolve(type);
+  }
+
   @Override
   public Statement apply(final Statement base, final Description description) {
     return new Statement() {


pettermahlen:
Hm, my gut reaction on seeing this was "brilliant!", but on thinking more about it, I'm not so sure. The reason is that the use of Guice, and hence the ability to resolve instances, is for apollo-internal things. We don't want to expose that externally and we should encourage people to use a separate Injector or DI system entirely for wiring up their application.

This PR would encourage people to reuse the internal injector for the application-level object graph.

That said, this does seem like a real need, and it would be great to support it somehow. Maybe a solution could be a way to get hold of the `AppInit` from the `ServiceHelper`, and then people can cast that/do whatever they want to get the objects they want from that? If people do use the internal Injector for their application wiring, they could even keep track of the `Environment` and use that in your use case above.

@rouzwawi you've thought more about the module system than I have, comments?

udoprog:
Hey!

@pettermahlen: The `<T> T resolve(Class<T> type)` doesn't expose anything specific to Guice though. And it's the same as exposed by `Service.Instance`.

zalenski:
> it's the same as exposed by  `Service.Instance`

Same as exposed in the public API `Environment.resolve(clazz)`.

pettermahlen:
@zalenski agreed, it does - but `Environment.resolve` allows you to access Apollo things from within your production code, whereas this would be specifically to access your own things from the Apollo Injector - first having added them to the Apollo Injector.

So the point is to not encourage using the Apollo modules system for application wiring. There's nothing explicit in the documentation or APIs about that, but it's something I think makes sense.

udoprog:
@pettermahlen I understand the want of not using it for application wiring, but I don't believe that this pull requests influences that behavior in any way.

The fundamental problem is that wee need to have access to Apollo module provided components both internally and externally through ServiceHelper for testing purposes, the `ApplicationModule` in the original post is not related to this.

rouzwawi:
@udoprog do you have examples of things you've had to have access to like this that are not things you've written yourself as part of your application? Trying to understand the non-application needs better.

As said before, the module system and DI capabilities it has is not meant to be used to wire application level components. What I've done in order to make it easy to wire my services differently in tests have been to write the service as I would do without Apollo; i.e. a main application class that takes a bunch of factory interfaces as arguments (some of them which might even require an `Environment` in their provide method). This class then has a `init(Environment)` method that I mount with the `ServiceHelper` or `HttpService` assemblies. That allows me to wire my application completely decoupled from what Apollo does, and at the same time allows part of my wiring to depend on stuff from `Environment` such as config or in some cases Apollo module components.

rouzwawi:
Also to add: application components that my test needs to interact with are either set up by the set itself (and injected into the application class), or exposed as part of the public interface of the application class.

udoprog:
We have a work in progress pubsub module which provides a fake in-memory implementation that can be used for integration-tests.

It's initialized like this in ServiceRunner:

```java
    StandaloneService
        .usingAppInit(new SomeApplication(), SERVICE_NAME)
        .withModule(new PubsubApolloModule())
        .build()
        .start(args);
```

The application fetches an instance of `PubsubContext` using the environment, which is provided by the module.

```
final PubsubContext context = environment.resolve(PubsubContext.class);
```

In tests, we do the same thing, but with ServiceHelper referencing the fake module.

```java
final ServiceHelper helper = ServiceHelper.create(new SomeApplication(), ServiceRunner.SERVICE_NAME)
        .withModule(new FakePubsubModule());

final PubsubContext context = helper.resolve(PubsubContext.class);
// ... do things with the context ...
```

I can add more indirection as proposed by @rouzwawi, but it seems redundant when the project already depend on all the service wiring that Apollo provides. Parts of the fake also depend on Apollo-specific injections (`@Named(Services.INJECT_SERVICE_NAME) String`) to provide better auditing.

pettermahlen:
Trying to distill this, it seems like the need is for integration testing of an Apollo Module?

It feels like we've got two or three developer roles that we're catering to: 

1. Framework developers - modifying the internals of the framework. (Not sure how relevant this category is)
1. Module developers - enriching the ecosystem by developing new modules and assemblies.
1. Application developers - using modules and assemblies to build services.

To my mind, `ServiceHelper` is/was geared entirely at application developers, but perhaps it should also be useful for module developers. It would be an interesting challenge to get a better separation of APIs for those two categories. A big difference is that framework and module developers need to be aware of things like Guice injection, whereas we'd like to keep application developers isolated from that.

I wonder if it's possible/a good pattern to subclass `ServiceHelper` and simply have a getter for the `PubsubContext`? Subclassing feels wrong, because composition is a better pattern - I'm sort of thinking that `ServiceHelper` is very closely related to an assembly as it needs to know the set of modules included and, when relevant, how to replace real implementations with fakes.

udoprog:
@pettermahlen `ServiceHelper` would be primarily used for application developers. They'd use the faked pubsub support to build integration tests for their service, not the module developer.

Exposing the resolve method is the cleanest way I can see module developers providing these fakes to application developers. I attempted to do what @rouzwawi suggested, but the end result is really messy and I end up with many lines of boilerplate that need to be repeated across projects. Extending `ServiceHelper` won't work at all if more than one module wishes to expose fakes.

All in all. This is a tiny change. It's targeted at testers. I think its useful enough to be allowed some space to be tried out in `1.x`, and if it proves problematic it can be revisited in a later major version. I can guarantee that it is very useful for the situation I'm addressing right now.

pettermahlen:
@udoprog have you seen our internal `SpotifyServiceHelper` in apollo-modules? That's what I'm talking about. You could perhaps create a class similar to:

```
public class PubsubServiceHelper {
   private static final FakePubsubModule MODULE = new FakePubsubModule();

   public static ServiceHelper addModules(ServiceHelper helper) {
     return helper.withModule(MODULE);
   }
   
   public static PubsubContext context() {
     return MODULE.context();
   }
}
```

If that is a path forward, it would remove the need for 'magic', making the origin of things much more explicit.

The suggested change is tiny, agreed, but the potential ramifications are very large. Basically, it means opening up all of Apollo's internals for access in tests, and even if it is possible to change in a major upgrade, if relying on the internals becomes common, the cost of that upgrade becomes very high.

udoprog:
The module in question uses dependencies provided by Apollo (config and service name). These cannot be accessed externally and before this patch we did what you suggested, but had to instantiate or provide these externally. Providing a fake module to install is much cleaner and more consistent.

Dependencies provided through the module system can't be considered internal. Changes to them would break third party modules and applications. They are part of the public API.

We do need to seriously consider locking down internal things if they are exposed. Guice's PrivateModule is an excellent option in that case.

pettermahlen:
I agree with 

> Dependencies provided through the module system can't be considered internal. Changes to them would break third party modules and applications. They are part of the public API.

That's something I'd like to move away from or make explicit - it's problematic that the public API is so obscure today. Hence my desire to minimise the exposure of injected things. 

I don't understand 
> Providing a fake module to install is much cleaner and more consistent.

I'm suggesting you should keep doing that - did you see the `PubsubServiceHelper` sketch above or the internal `SpotifyServiceHelper` class in apollo-modules?

udoprog:
To help you understand my use-case, this is a simplified view of what the fake module currently looks like:

```
public class FakePubsubModule extends PrivateModule implements ApolloModule {
  private final FakePubsubEnvironment environment;

  @Provides
  @Named("service")
  public String publisher(final @Named(Services.INJECT_SERVICE_NAME) String serviceName) {
    return serviceName;
  }

  @Override
  protected void configure() {
    bind(FakePubsubEnvironment.class).toInstance(environment);

    bind(PubsubContext.class).to(FakePubsubContext.class).in(Scopes.SINGLETON);

    expose(PubsubContext.class);
    expose(FakePubsubEnvironment.class);
  }
```

You can see that there is a local dependency to `@Named(Services.INJECT_SERVICE_NAME) String`. The propose helper would therefore have to pass in this. We also have a similar dependency to access the `Config` as instantiated by Apollo, the stanza we had before resolve was to in the testcase create an instance of Config, and provide the ServiceRunner.SERVICE_NAME constant when building the module.

This result in the setUp of the test-case looking something like this.

```java
public class MyApplicationIT {
  ServiceHelper helper;

  PubsubContext context;

  @Before
  public void setUp() {
    final FakePubsubEnvironment environment = ...;
    final Config config = /* somehow, we basically copied what Apollo was doing */;

    final FakePubsubHelper fakePubsubHelper = new FakePubsubHelper(ServiceRunner.SERVICE_NAME, config, environment);

    helper = ServiceHelper.create(new Application(), "application")
        .withModule(fakePubsubHelper);

    context = fakePubsubHelper.context();
  }
}
```

Instead of this.

```java
public class MyApplicationIT {
  ServiceHelper helper;
  PubsubContext context;

  @Before
  public void setUp() {
    final FakePubsubEnvironment environment = ...;

    helper = ServiceHelper.create(new Application(), "application")
        .withModule(new FakePubsubModule(environment));

    context = helper.resolve(PubsubContext.class);
  }
}
```

Now consider every dependency that your application has, some of them are useful in the test as well. Instead of introducing more indirection through helpers, the module system can be used to consistently provide instances of these. And at least for extension modules this makes sense.

I'll provide you with a link OOB to showcase what happened with our code when we made this transition.

rouzwawi:
@udoprog, My suggestion doesn't require you to create local instances of those Apollo managed objects. You're supposed to get them through `Environment`.

I think the boilerplate it adds should already be scaffolding that's set up for making the service testable. Like, the pubsub dependency is likely just another thing among others that need to be injected.

udoprog:
@rouzwawi I started by doing it your way. This change allows doing it with less boilerplate for the test cases.

I propose this change to remove the need for the Helper indirection. I can now build a module that can be used in place of the real one. And resolve helper methods from the ServiceHelper.

You also have not addressed the components depending on things provided by Apollo through the module system. In a general sense, not just a few magic strings or things that can be built using factory methods.
Try this; build a class that depends on `Closer` or `Environment` (for `Config`) that is both provided in the application, and is available to the test. I know it's possible, but it's so straight forward for everyone if I can just build that instance from the module system.

udoprog:
I've created a gist to showcase a canonical example: https://gist.github.com/udoprog/7493bb2fae15c8ab202b36d0cad3291f

rouzwawi:
@udoprog I posted a comment with the setup I had in mind.

rouzwawi:
Moving comment here

> btw, for this to get something out from within Apollo, the Environment.resolve(Class) method is available. But again, I'd like stuff to not unnecessarily be tied to Apollo.

> The setup I suggest could easily be replaced with something like Dagger2 that would solve the problem of wiring multiple dependencies and managing singletons, etc.

udoprog:
The setup you propose have the same amount of repeated boilerplate as the example I proposed. I have no idea why you posted it since It's just another way of accomplishing what I set out to simplify. I've seen 10 different ways of doing that. The point is not that it can be done, but that it _has to be done_ again and again in every single project.

This PR proposes to unify the approach for module developers to provide modules both for production and testing. That can have complex or simple dependencies without causing the application developer to bend over backwards making up their own novel ways of accessing the environment in their tests.

rouzwawi:
One reason I posted it was that your `WithoutResolve` example still used an Apollo Module for the PubSubModule. That lead to the `Optional<PubSub>` field and and the non-uniformity of how to set up the application. It suffers from throwing runtime exceptions when the user forgets to set up the right combination of modules.

I wanted to show that you'd have a much easier to understand (from the users perspective) way of obtaining the `PubSub` object without having to even know that Apollo has a module system.

As to saying that it's boilerplate, I have to disagree. It's just regular Java code for making things testable. I don't want Apollo to be a framework that everything has to be integrated with. Rather I want it to just be a library among others. So if you'd remove Apollo from the equation, you'd still have the same code there for wiring the application.

So for anything to be a module, I'd ask "Why does it have to be a module?". I there's no good reason then it doesn't have to be a module. And that it needs to access `Environment` or the `Closer` are not good reasons.

udoprog:
I can buy that you don't want it to be a module. That does make it an issue that a lot of things are supported through the use of modules. A module is probably also what a good chunk of users would expect in 1.x. Up until now I've assumed this is the de facto standard of extending the functionality of Apollo apps and I might be mistaken.

But now it literally is boilerplate to having to invent ways of injecting and exposing certain things in each application to support tests. And there is an alternative.

rouzwawi:
So I'm thinking that this comes down to expectations and communication around how to add functionality to apps.

I've held the position that things should be done outside of modules, but I realise that has not been what seems to be the general idea among other users. Perhaps a separate discussion, but it would be good to get some wider input on what strategy we should be following.

pettermahlen:
I think a fundamental issue is the lack of clarity on what we want from the module system. In principle, everything can be done without it, including things like setting up HTTP servers, managed clients, etc., for which we're today providing modules and assemblies that configure combinations of them. So what is it that decides whether access to Pub/Sub is a good candidate for a module or not? I don't know.

Another problem I have with the module system, as mentioned above, is the lack of clarity on how things are wired together/the lack of clear APIs of what's available. That's illustrated by the 'magic' nature of `Environment.resolve()`. It's very hard to determine if an attempt at resolving a class will work or not from looking at static code.

I think that's a reason why I'm reluctant to expand on modules as an important thing. I'd probably rather work on retiring them in some future.

udoprog:
I don't really want to get into the discussion what you want to do with modules. But as long as Environment has a resolve it's odd that ServiceHelper doesn't. It causes the kind of pseudo solutions you see in my gist.

udoprog:
So this PR is currently left in suspended state. I either really need this to support the my current module-centric approach I'm exposing for testing and interacting with Pub/Sub, or I need to rewrite my module and other things depending on it to conform with the other approach. When or how can I expect a decision?

lepistone:
Closing because of inactivity.

Issue #156, opened 2016-08-22
Incorrect request logging on timeout


pettermahlen:
The following may happen:

1. A request arrives, and processing starts
1. The TTL expires, and a 500 Timeout response is sent
1. Processing finishes successfully, and the request is logged as 200 OK (or something else)

pettermahlen:
This would also affect statistics collected. A couple of possible solutions:

1. Change `OngoingRequest.reply(Response<ByteString>)` to (optionally?) return the actual response sent, permitting inner requests to change it and communicate those changes back out the decorator chain. This would allow the logger to understand both that there was a previous timeout, and that the request eventually succeeded.
1. Make reporting request outcomes a primary concern of the innermost request handler, allowing both metrics and logging to be handled via something analogous to the `OutcomeReportingOngoingRequest`, after the point that the final/real response is sent. In this case, the inner request needs to handle multiple attempts at replying - which it probably needs today as well.

pettermahlen:
@nabam does this description seem correct?

nabam:
@pettermahlen It actually seems like nothing is logged at all on timeout.

pettermahlen:
Summary of @nabam's observations: 

1. A service with a very low timeout setting, ensuring the request times out:
  1. The response returned is the timeout that is expected
  1. A custom logger logs 'success' when the work completes (after the timeout)
  1. Nothing gets logged by the default access logger
1. The same service, with a higher timeout setting, ensuring the request completes:
  1. The successful response is returned.
  1. The custom logger logs 'success'
  1. The default access logger logs 'success'

Ideal behaviour in the timeout case would be:

1. The timeout gets logged (and reported in statistics, etc.) correctly
1. The ongoing work gets cancelled

PR #157, opened 2016-08-24
ensure failures to respond are logged

pettermahlen:
Prior to this change, a failure to send a response (for an error or a correctly
handled request) to clients would remain 'sitting' in a failed CompletionStage.

Partial remediation for #156


jo-ri:
+1

PR #161, opened 2016-08-29
Improve outcome logging

pettermahlen:
The current, decorator-based, solution for logging request outcomes suffers from 
decorators' ability to modify the actual response sent. An example of this is when 
the `AsyncContextOngoingRequest` request processing gets short-circuited by a Jetty
timeout, but it can happen in other ways as well. To solve this, this PR makes request 
logging a concern of the innermost `OngoingRequest` rather than a decorator concern.

The same is applicable to request statistics; a timeout today will not be correctly metered. That is intentionally left out of this PR, to be fixed in 2.0 only, instead.

### DIFF
--- a/apollo-extra/README.md
+++ b/apollo-extra/README.md
@@ -20,6 +20,12 @@ Contains some serializer middlewares, and utilities for versioning endpoints.
 
 ## com.spotify.apollo.logging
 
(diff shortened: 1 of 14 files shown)


jo-ri:
+1 Nice!

PR #167, opened 2016-09-21

### LINKED ISSUE #129: How to get the Request protocol from Request.uri()?
When I made a request and try to capture the Request protocol, I can't get it from Request.url().  How to do get it?

According to the W3 doc, it should have it from the Reqest-URI
https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html

Example:
GET http://www.w3.org/pub/WWW/TheProject.html HTTP/1.1

How do I get the "HTTP/1.1"?

### LINKED ISSUE #148: Ability to access network layer information in HTTP request context
Currently, there is no way to access network and transport layer information such as client/server IPs and ports in HTTP request context. In some cases, this information is necessary for example for GeoIP checks or whitelisting. Having such ability would be nice.

pettermahlen:
Adds a facility allowing server implementations to annotate
incoming requests with metadata that can then be used in
request handlers. Also adds an implementation in HttpService
that provides the metadata that has been requested to date.

Fixes #129
Fixes #148

### DIFF
--- a/apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
@@ -1,0 +1,65 @@
+/*
+ * -\-\-
+ * Spotify Apollo API Interfaces
+ * --
+ * Copyright (C) 2013 - 2016 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+package com.spotify.apollo;
+
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+
+/**
+ * Describes an API for retrieving metadata about an incoming request.
+ */
+public interface RequestMetadata {
+
+  String METADATA_SOURCE = "apollo.metadata-source";
+
+  /**
+   * Get the arrival time of the incoming request in nanoseconds. Note that this is not
+   * unix epoch as the time is provided by {@link System#nanoTime()}. To get unix epoch
+   * time, do something like:
+   *
+   * <pre>
+   * {@code
+   * long processingTimeNanos = System.nanoTime() - requestContext.arrivalTimeNanos();
+   * long arrivalTimeUnixEpochMillis = System.currentTimeMillis() +
+   *                                   TimeUnit.NANOSECONDS.toMillis(processingTimeNanos);
+   * }
+   * </pre>
+   *
+   * @see System#nanoTime()
+   */
+  default long arrivalTimeNanos() {
+    // This is not a good default for real implementations. It is simply a catch-all
+    // default to not break existing implementations.
+    return System.nanoTime();
+  }
+
+  /**
+   * Returns metadata for this request; may include things like information about the origin of the
+   * request (remote IP address of the connected socket), etc. The exact content is server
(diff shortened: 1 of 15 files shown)


modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java
```diff
@@ -88,6 +96,15 @@ public void handle(
     baseRequest.setHandled(true);
   }
 
+  private Map<String, String> extractMetadata(HttpServletRequest req) {
+    return ImmutableMap.of(
+        JettyHttpRequestMetadata.PROTOCOL_VERSION.name(), req.getProtocol(),
+        JettyHttpRequestMetadata.REMOTE_ADDRESS.name(), req.getRemoteAddr(),
+        JettyHttpRequestMetadata.REMOTE_PORT.name(), String.valueOf(req.getRemotePort()),
+        RequestMetadata.METADATA_SOURCE, this.getClass().getName()
```
jo-ri: Can request metadata only be added in one place?

modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/JettyHttpRequestMetadata.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ * -/-/-
+ */
+package com.spotify.apollo.http.server;
+
+/**
+ * Defines the metadata reported for {@link com.spotify.apollo.RequestContext}s originating from
+ * this server.
+ */
+public enum JettyHttpRequestMetadata {
```
jo-ri: Must client be aware that Jetty is used to read these metadata properties?

modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/JettyHttpRequestMetadata.java
```diff
@@ -0,0 +1,30 @@
...
+  PROTOCOL_VERSION,
+  REMOTE_ADDRESS,
+  REMOTE_PORT
```
jo-ri: I think these are general enough that I would prefer explicit methods instead of String -> String map (maybe combining addess:port)

apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Returns metadata for this request; may include things like information about the origin of the
+   * request (remote IP address of the connected socket), etc. The exact content is server
+   * implementation-dependent. Implementations are encouraged to clearly document which set of keys
+   * and values they will record as request metadata. All implementations are encouraged to set
+   * a value for the key {@link #METADATA_SOURCE}, indicating a Java class that generated the
+   * metadata; this enables users to more easily understand what data to expect.
+   */
+  default Map<String, String> metadata() {
```
pettermahlen: That should be safe, `OngoingRequest` is a 'very internal' interface (which wouldn't be exported if Apollo was built on Jigsaw), and it's not likely at all that people will implement `RequestContext` either.

modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java
pettermahlen: No, it could be added anywhere - in the Jetty implementation, it's only added in one place for now.

modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/JettyHttpRequestMetadata.java
pettermahlen: The purpose of this enum is to document what metadata gets added by the Jetty HTTP server. The client can iterate through the contents of the metadata map, so the data can be read without knowing about this enum.

modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/JettyHttpRequestMetadata.java
pettermahlen: So, another method in `RequestMetadata`, analogous to `arrivalTimeNanos()`?

pettermahlen:
Closing in favour of #169

PR #168, opened 2016-09-22

### LINKED ISSUE #129: How to get the Request protocol from Request.uri()?

### LINKED ISSUE #148: Ability to access network layer information in HTTP request context

pettermahlen:
Adds a facility allowing server implementations to annotate
incoming requests with metadata that can then be used in
request handlers. Also adds an implementation in HttpService
that provides the metadata that has been requested to date.

This is an alternative to #167 

Fixes #129 
Fixes #148


apollo-api/src/main/java/com/spotify/apollo/RequestContext.java
```diff
@@ -52,23 +52,7 @@
...
-   * long processingTimeNanos = System.nanoTime() - requestContext.arrivalTimeNanos();
-   * long arrivalTimeUnixEpochMillis = System.currentTimeMillis() +
-   *                                   TimeUnit.NANOSECONDS.toMillis(processingTimeNanos);
-   * }
-   * </pre>
-   *
-   * @see System#nanoTime()
+   * Returns the metadata available for this request.
    */
-  default long arrivalTimeNanos() {
```
jo-ri: Deprecate but keep and implement as metadata().arrivalTimeNanos() to ease migration?

apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Describes an API for retrieving metadata about an incoming request.
+ */
+public interface RequestMetadata {
+
+  /**
+   * Returns the class that is the source of the metadata; this enables users to more easily
+   * understand where the metadata comes from and how to interpret data like the {@link #protocol()}.
+   */
+  Class<?> source();
```
jo-ri: Implies that request metadata can only be added in one place. Is that necessary?

apollo-api/src/main/java/com/spotify/apollo/RequestContext.java
pettermahlen: I would prefer to simplify/remove this rather than deprecate because:

- it's a part of a major version upgrade, 
- it's an easy change to make, and
- it's not normally used by application code, but only by framework-type code (like our Hermes implementation)

apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
pettermahlen: Hmm... With the current set of metadata fields, we will only need to add them in one place. With the solution in #167, that has a general map of metadata properties, there might be value in allowing 'appending' to the metadata properties in more than one place.

The reason for using a `Class<?>` rather than a `String` as the source name is to facilitate tracking down where the metadata was added. It makes it harder for implementors to provide a bad/hard to find source.

apollo-api/src/main/java/com/spotify/apollo/RequestContext.java
jo-ri: Ok, great!

apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
jo-ri: Seems implementation dependant if all metadata properties are added in the same place or not. Even more so if we add more properties to RequestMetadata. If client really need to depend on the creator of the metadata, maybe RequestMetadata could be implemented by a class that the client can downcast to?

apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
```diff
@@ -0,0 +1,70 @@
...
+   * {@code
+   * long processingTimeNanos = System.nanoTime() - requestContext.arrivalTimeNanos();
+   * long arrivalTimeUnixEpochMillis = System.currentTimeMillis() +
+   *                                   TimeUnit.NANOSECONDS.toMillis(processingTimeNanos);
+   * }
+   * </pre>
+   *
+   * @see System#nanoTime()
+   */
+  long arrivalTimeNanos();
```
jo-ri: Return java.time.Instant instead?

apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
pettermahlen: I don't mean for client *code* to depend on the creator, I mean for *developers* to be able to easily figure out where the data comes from, in order to understand it. The 'source' is 'just' documentation.

apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
pettermahlen: yes!

pettermahlen:
Closing in favour of #169

PR #169, opened 2016-09-22

pettermahlen:
Yet another alternative to #167 and #168.


apollo-api/src/main/java/com/spotify/apollo/RequestContext.java
```diff
@@ -65,10 +67,35 @@
...
   default long arrivalTimeNanos() {
     // This is not a good default for real implementations. It is simply a catch-all
     // default to not break existing implementations.
     return System.nanoTime();
   }
+
+  /**
+   * Returns the metadata available for this request.
+   */
+  default RequestMetadata metadata() {
```
jo-ri: Since this technically can break builds, maybe we should bump the minor version and mention it in the release note?

apollo-api/src/main/java/com/spotify/apollo/RequestMetadata.java
```diff
@@ -0,0 +1,60 @@
...
+   * Indicates the remote address of the connection which the request was received on, if available.
+   * This may be the address of a proxy address or a direct connection to the caller, depending on
+   * the network setup.
+   */
+  Optional<HostAndPort> remoteAddress();
+
+  /**
+   * Defines an address
+   */
+  interface HostAndPort {
```
pettermahlen: I actively made the choice to use a simpler data type, since InetSocketAddress has this DNS resolution logic in it.

apollo-api/src/main/java/com/spotify/apollo/RequestContext.java
pettermahlen: Yes, good point

PR #173, opened 2016-10-18
use FOSS root pom

pettermahlen:
Clean up some stuff from the parent POM in this project.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,12 @@
     <name>Spotify Apollo Parent</name>
     <url>https://github.com/spotify/apollo</url>
 
-    <groupId>com.spotify</groupId>
+    <parent>
+        <groupId>com.spotify</groupId>
+        <artifactId>foss-root</artifactId>
+        <version>3</version>
+    </parent>
+
     <artifactId>apollo-parent</artifactId>
     <version>1.3.1-SNAPSHOT</version>
     <packaging>pom</packaging>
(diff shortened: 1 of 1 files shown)


pettermahlen:
@rouzwawi look good?

rouzwawi:
Nice. Does the failsafe plugin get bound by default? The same with enforcer. Seems some plugins were completely removed. Maybe that was intentional.

pettermahlen:
The build section of the pom still has this: 
```
        <plugins>
            <plugin>
                <artifactId>maven-enforcer-plugin</artifactId>
            </plugin>
            <plugin>
                <artifactId>maven-surefire-plugin</artifactId>
            </plugin>
            <plugin>
                <artifactId>maven-failsafe-plugin</artifactId>
            </plugin>

            <plugin>
                <groupId>org.eluder.coveralls</groupId>
                <artifactId>coveralls-maven-plugin</artifactId>
            </plugin>
            <plugin>
                <groupId>org.codehaus.mojo</groupId>
                <artifactId>license-maven-plugin</artifactId>
            </plugin>
        </plugins>
    </build>
```

When I run `mvn verify`, they both run, so that looks OK to me.

pettermahlen:
I guess we could remove the coveralls plugin though?

rouzwawi:
Aha I missed that.

Isn't coveralls running?

pettermahlen:
It might be, but we're using codecov.io these days, so AFAIK, coveralls no longer fulfils a need?

rouzwawi:
That's right. Confused those two. This is way too "early" in the morning for me :)

PR #180, opened 2016-11-16
clarify unimplemented RequestContext metadata

pettermahlen:
Previously, if users upgrading do not notice that there is a new
RequestContext.metadata() method, their code will start failing
in strange ways (requests timing out was the concrete example)
because of the values supplied by the default implementation. It's
a better solution to throw an exception, forcing implementations
to actually override this method.


jwiklund:
:+1:

PR #181, opened 2016-11-16
clarify unimplemented RequestContext metadata

pettermahlen:
Previously, if users upgrading do not notice that there is a new
RequestContext.metadata() method, their code will start failing
in strange ways (requests timing out was the concrete example)
because of the values supplied by the default implementation. It's
a better solution to throw an exception, forcing implementations
to actually override this method.

### DIFF
--- a/apollo-api/src/main/java/com/spotify/apollo/RequestContext.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/RequestContext.java
@@ -81,21 +81,6 @@ public interface RequestContext {
    * Returns the metadata available for this request.
    */
   default RequestMetadata metadata() {
-    return new RequestMetadata() {
-      @Override
-      public Instant arrivalTime() {
-        return Instant.EPOCH;
-      }
-
-      @Override
-      public Optional<HostAndPort> localAddress() {
-        return Optional.empty();
-      }
-
-      @Override
-      public Optional<HostAndPort> remoteAddress() {
-        return Optional.empty();
-      }
-    };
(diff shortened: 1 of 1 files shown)


udoprog:
This caught my attention and I wanted to investigate what would happen if you'd just remove the defender method completely. It opened my eye to another compatibility concern regarding defender methods and I ended up doing some experiments to verify my assumptions.

While this is part of the public API, consumers won't notice as long as an implementation is present. Assuming all implementations have overridden it so far by convention, a mismatch in API version (more recent than impl) would also be safe. `default` ended up not being part of the method signature so that method dispatch will work as before for binary compatibility whether it is present or not. What would happen is if a previous implementation does not override the method and the API is updated independently, an `AbstractMethodError` would be raised. Which is arguably worse than a `RuntimeException`.

pettermahlen:
@udoprog that's sort of what I was thinking, too, when I realised that this default implementation really didn't help users, as it was broken. So probably, this change (adding metadata and giving that  meaning in implementation) should really have been part of a major version upgrade, I guess.

The current implementation isn't really any better, I think, than getting a runtime error due to a missing method. In either case, you need to trigger that code path in order to detect the problem, instead of getting help from the compiler. My bad, but at least I learned something, I hope.

pettermahlen:
And I didn't notice until now that this PR is for master, which is anyway backwards-incompatible. So rather than replicating what was done in #180 for 1.x, for v2, we should just remove the default implementation.

PR #183, opened 2016-11-30
ensure FakeOngoingRequest uses a better arrival timestamp

pettermahlen:
Without this, tests using the ServiceHelper and sending real requests get incorrect
values for arrivalTime, leading to broken request handling.


jappaleinen-sp:
👍

PR #186, opened 2016-12-20
Remove .txt from LICENSE and NOTICE

davidxia:
According to Apache license spec, these files should
not have an extension.


rouzwawi:
Thanks!

PR #187, opened 2016-12-21
Make comparing status more convenient.

rydenius:
More than once I have seen code like this fail:
```
if (Status.OK == response.status()) {
```
or
```
if (Status.OK.equals(response.status()) {
```
because the status of the response has a different reason phrase. The current solution would be to use
```
if (Status.OK.code() == response.status().code()) {
```
This small patch would simplify this check and also signal to developers that regular equals method might not be what you want to use.

### DIFF
--- a/apollo-api/src/main/java/com/spotify/apollo/StatusType.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/StatusType.java
@@ -45,7 +45,7 @@ public interface StatusType {
    * @return true if code of this equals code of statusType
    */
   default boolean equalCode(StatusType statusType) {
-    return statusType != null && code() == statusType.code();
+    return code() == statusType.code();
   }
 
   /**
@@ -55,7 +55,7 @@ public interface StatusType {
    * @return true if family of this equals family of statusType
    */
   default boolean equalFamily(StatusType statusType) {
-    return statusType != null && family() == statusType.family();
+    return family() == statusType.family();
   }
 
   /**
--- a/apollo-api/src/main/java/com/spotify/apollo/StatusType.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/StatusType.java
@@ -39,6 +39,26 @@ public interface StatusType {
   StatusType withReasonPhrase(String reasonPhrase);
 
   /**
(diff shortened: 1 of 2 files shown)


danielstahl:
👍

martina-if:
:+1:

bridell:
👍

mattnworb:
👍  I've made this mistake often

apollo-api/src/main/java/com/spotify/apollo/StatusType.java
```diff
@@ -39,6 +39,26 @@
   StatusType withReasonPhrase(String reasonPhrase);
 
   /**
+   * Convenience method for comparing the code of two StatusType instances.
+   *
+   * @param statusType the instance to compare this to
+   * @return true if code of this equals code of statusType
+   */
+  default boolean equalCode(StatusType statusType) {
+    return statusType != null && code() == statusType.code();
```
dflemstr: Would rather have this throw NPE... Fail early!

apollo-api/src/main/java/com/spotify/apollo/StatusType.java
```diff
@@ -39,6 +39,26 @@
...
+  }
+
+  /**
+   * Convenience method for comparing the family of two StatusType instances.
+   *
+   * @param statusType the instance to compare this to
+   * @return true if family of this equals family of statusType
+   */
+  default boolean equalFamily(StatusType statusType) {
+    return statusType != null && family() == statusType.family();
```
dflemstr: Same

apollo-api/src/main/java/com/spotify/apollo/StatusType.java
pettermahlen: I would agree, I think. @rydenius, is there a reason to expect/permit users to pass in a null `statusType` to this method?

rydenius: Fixed now.

apollo-api/src/main/java/com/spotify/apollo/StatusType.java
rydenius: Fixed now.

PR #190, opened 2017-02-02
make CombinedFormatLogger public

mattnworb:
Allow someone who wants to change the RequestOutcomeLogger used by
HttpServerProvider to customize the default logging behavior by chaining
multiple consumers together (since RequestOutcomeLogger is a BiConsumer,
and therefore has an `andThen(..)` method).

For example: combining the default logging behavior with an additional
RequestOutcomeLogger which would emit something about the request logs
to logstash/elasticsearch, but without having to reinvent the
CombinedFormatLogger behavior (since it is non-public today).


mattnworb:
@rouzwawi @pettermahlen

pettermahlen:
I'm not in a position to check the code right now, but the one objection I can think of is that it doesn't feel quite right for the jetty-http-server to have this logger as part of its API. The server should be a server, not a utilities library. Would it be easy to move this class somewhere else? apollo-extra? Somewhere else?

lepistone:
Closing because of inactivity. Feel free to re-open if appropriate.

PR #192, opened 2017-02-17
Add method to StubClient verifying mappings

eshrubs:
This is a useful in unit tests to ensure that all requests were sent by
the stub client. In the case that a Request status isn't important, but
sending it is, such as a "fire-and-forget" scenario, this is nice to verify.

For example,

```java
stubClient.addMapping(uri("http://fire-and-forget"), ...);

...
// Ensure that all mappings sent
stubClient.verifyAllMappingsSent() // will assert everything was sent.
```

Previously this was achievable but was hacky

```java
stubClient.addMapping(uri("http://fire-and-forget"), ...);

...
// Ensure that all mappings sent
assertThat(stubClient.sentRequests(), contains(uri("http://fire-and-forget"));
// or...
assertThat(stubClient.sentRequests(), size(1));
```

This verification is simple, and does not handle sequenced ResponseSources
or checks that a match was made multiple times. But it is a good first step.

### DIFF
--- a/apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
+++ b/apollo-test/src/main/java/com/spotify/apollo/test/StubClient.java
@@ -21,7 +21,6 @@ package com.spotify.apollo.test;
 
 import com.google.auto.value.AutoValue;
 import com.google.common.base.Preconditions;
-
 import com.spotify.apollo.Client;
 import com.spotify.apollo.Request;
 import com.spotify.apollo.Response;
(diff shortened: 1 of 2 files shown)


martina-if:
@eshrubs you need to duplicate this PR against the branch 1.x if you want it released before apollo 2.0, since master branch is version 2.

martina-if:
:+1:

eshrubs:
@martina-if I will make a bulk PR against 1.x with #193 #192 #191 #188

PR #193, opened 2017-02-17
StubClient more details on no match

eshrubs:
Today you don't get much details on what was not matched when
No match was found. However this can be confusing in testing,
as you never see what the cause of the mismatch was without debugging 
or trial/error.

This will detail the expected matchers.


rschildmeijer:
👍

martina-if:
:+1:

martina-if:
@eshrubs if you want this released on the next 1.x release you need to copy this PR against the branch 1.x

eshrubs:
@martina-if Ok I will add a PR against 1.x

eshrubs:
@martina-if I will make a bulk cherry-pick against 1.x when these have been merged to 2.x

PR #194, opened 2017-02-21
Log remote ip address for incoming requests

martina-if:
@jori @ada

### DIFF
--- a/modules/jetty-http-server/pom.xml
+++ b/modules/jetty-http-server/pom.xml
@@ -66,11 +66,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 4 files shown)


modules/jetty-http-server/pom.xml
```diff
@@ -66,11 +66,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-            <groupId>ch.qos.logback</groupId>
-            <artifactId>logback-classic</artifactId>
```
martina-if: I removed logback-classic because we already had a logging binding from slf4j-test

jo-ri:
👍

PR #195, opened 2017-02-28
Add a minimum for reply rate in error ratio calculation

johanburatti:
This fixes an issue where the error-reply-ratio metric will show un-predictable behavior after sites or hosts have been drained without the hosts being restart afterwards.

### DIFF
--- a/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
+++ b/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
@@ -50,10 +50,14 @@ import static com.spotify.apollo.metrics.semantic.What.ERROR_RATIO;
 import static com.spotify.apollo.metrics.semantic.What.REQUEST_FANOUT_FACTOR;
 import static com.spotify.apollo.metrics.semantic.What.REQUEST_PAYLOAD_SIZE;
 import static com.spotify.apollo.metrics.semantic.What.RESPONSE_PAYLOAD_SIZE;
+import static java.lang.Math.max;
 import static java.util.Objects.requireNonNull;
 
 @SuppressWarnings("OptionalUsedAsFieldOrParameterType")
 class SemanticServiceMetrics implements ServiceMetrics {
+
+  // Minimum reply rate at which we consider the error ratio to make sense.
+  private static final double ERROR_GAUGE_MINIMUM_REPLY_RATE = 1e-3;
 
   private final SemanticMetricRegistry metricRegistry;
   private final MetricId metricId;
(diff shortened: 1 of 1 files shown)


modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
```diff
@@ -203,6 +207,18 @@ protected Ratio getRatio() {
         });
   }
 
+  private Supplier<Ratio> errorRatioSupplier(Supplier<Double> errorRateSupplier,
+                                             Supplier<Double> replyRateSupplier) {
+    // If a service is drained of traffic (for instance by moving traffic to another host or site),
+    // the meter values will decay exponentially towards zero. If left for some time (hours),
+    // rounding errors will cause the values of the meters to approach each other, thus skewing
```
mattnworb: you were seeing the gauge value skew up towards 1 in these cases?

johanburatti: Exactly. In most cases it makes a jump from almost zero to almost 1. Has woken up people several times already. Do you want me to clarify the comment even more?

mattnworb: I think it is clear but just wanted to make sure I understood correctly 👍

kouKatsumi:
@johanburatti 
The change looks reasonable. However I'm not  very familiar with metrics code. Is there a way to easily test it? If yes, could you add tests?

johanburatti:
@kouKatsumi it is quite hard to test since it depends on timing and floating point rounding. We already have tests that cover the code. I could write one more test that looks like it does something clever, but it would likely not help us prevent any bugs. Do you still want the test?

stefanalund:
@johanburatti you had been running the code in production, right? Not saying that means we don't need tests, but at least gives some indicator of whether it's mergeable.

stefanalund:
@martina-if WDYT?

johanburatti:
@stefanalund no this has not run in production because we need to merge it before we can use it in production, unless we fork apollo completely. But this change is very small. One way is to merge it and then use the snapshot build as a dependency. I think that is the easiest way here.

johanburatti:
@stefanalund I also want to point out again, that we do have tests that cover the code. We just do not have any tests that will be able to measure the difference. This is because apollo does not allow test code to mock the meters. The meters in turn do not allow test code to mock system time. Since the change in behavior is only visible after a couple of hours without requests it is not feasible to create a test for this change. But again, we already _do_ test that the ratio behaves as expected in the unit tests.

martina-if:
:+1:

martina-if:
@johanburatti I think this PR is against master, which means, it won't be released until we release 2.0. Can you replicate it for the branch 1.x?

johanburatti:
@martina-if Oh, I did not notice that master was aiming for 2.0. I have made a new PR for 1.x here: #201.

johanburatti:
Closing this as it was merged to the 1.x branch instead

martina-if:
I think it has to be merged in both branches

martina-if:
@johanburatti is this the ~exact same code as for the PR to 1.x? In that case I can merge this as well so we have the same features in both branches

PR #199, opened 2017-04-18
Look up headers in a case insensitive way

martina-if:
In this first step apollo will look up headers in a case insensitive way but it will still preserve the original letter case of header fields when propagated. This makes services compatible with services that send lowercase headers (like those who propagate headers from an HTTP 2 connection). 

I recommend looking at the commits separately.

@pettermahlen @rouzwawi

### DIFF
--- a/apollo-api/src/main/java/com/spotify/apollo/Headers.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/Headers.java
@@ -1,0 +1,82 @@
+/*
+ * -\-\-
+ * Spotify Apollo API Interfaces
+ * --
+ * Copyright (C) 2013 - 2017 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+package com.spotify.apollo;
+
+import com.google.auto.value.AutoValue;
+import com.google.common.base.Preconditions;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.AbstractMap.SimpleImmutableEntry;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+@AutoValue
+abstract class Headers {
+
+  static Headers EMPTY = create(Collections.emptyMap());
+
+  static Headers create(Map<String, String> headers) {
+    final List<Map.Entry<String, String>> headersList = new ArrayList<>(headers.size());
+    headers.entrySet().forEach(h -> insertIfAbsent(headersList, h));
+    return new AutoValue_Headers(ImmutableList.copyOf(headersList));
+  }
+
+  public Optional<String> get(String name) {
+    Preconditions.checkArgument(name != null, "Header names cannot be null");
+
+    for (int i = 0; i < entries().size(); i++) {
+      final Map.Entry<String, String> headerEntry = entries().get(i);
+      if (name.equalsIgnoreCase(headerEntry.getKey())) {
+        return Optional.ofNullable(headerEntry.getValue());
+      }
+    }
+
+    return Optional.empty();
+  }
+
+  public Map<String, String> asMap() {
+    ImmutableMap.Builder<String, String> headers = ImmutableMap.builder();
+    entries().forEach(headers::put);
+    return headers.build();
+  }
+
+  public abstract ImmutableList<Map.Entry<String, String>> entries();
+
+  private static void insertIfAbsent(List<Map.Entry<String, String>> headerList,
+                                     Map.Entry<String, String> newHeader) {
+
+    // Replace existing header with new key (letter case can be overwritten) and value
+    for (int i = 0; i < headerList.size(); i++) {
+      Map.Entry<String, String> currentHeader = headerList.get(i);
+      if (currentHeader.getKey().equalsIgnoreCase(newHeader.getKey())) {
+        headerList.set(i, new SimpleImmutableEntry<>(newHeader));
+        return;
+      }
+    }
+
+    // No matching entry present, add new entry
+    headerList.add(new SimpleImmutableEntry<>(newHeader));
+  }
+
+}
--- a/apollo-api/src/main/java/com/spotify/apollo/Headers.java
+++ b/apollo-api/src/main/java/com/spotify/apollo/Headers.java
@@ -20,7 +20,6 @@
 package com.spotify.apollo;
 
 import com.google.auto.value.AutoValue;
-import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 
@@ -29,6 +28,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 @AutoValue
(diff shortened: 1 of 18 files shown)


martina-if:
Maybe it's not worth having Headers being an AutoValue type, what do you think?

apollo-api/src/main/java/com/spotify/apollo/Headers.java
```diff
@@ -0,0 +1,82 @@
...
+  static Headers EMPTY = create(Collections.emptyMap());
+
+  static Headers create(Map<String, String> headers) {
+    final List<Map.Entry<String, String>> headersList = new ArrayList<>(headers.size());
+    headers.entrySet().forEach(h -> insertIfAbsent(headersList, h));
+    return new AutoValue_Headers(ImmutableList.copyOf(headersList));
+  }
+
+  public Optional<String> get(String name) {
+    Preconditions.checkArgument(name != null, "Header names cannot be null");
```
pettermahlen: I think it's better to use `Objects.requireNonNull` in this case - generally null arguments lead to NPEs, and it's better to use the JDK classes than Guava where feasible.

apollo-api/src/main/java/com/spotify/apollo/Request.java
```diff
@@ -67,14 +69,31 @@
...
+  Iterable<Map.Entry<String, String>> headerEntries();
+
+
+  /**
+   * A header of the request message, looked up in a case insensitive way,
+   * or empty if no header with that name is found.
    */
   default Optional<String> header(String name) {
-    return Optional.ofNullable(headers().get(name));
+    Preconditions.checkArgument(name != null, "Header names cannot be null");
```
pettermahlen: Would it be better to permit nulls here, and reverse the comparison below so that it would just fail to find something, rather than throw an exception for a null? I'm not sure. Either way, if nulls shouldn't be permitted, then I think we should use `Objects.requireNonNull` for the argument check.

apollo-api/src/main/java/com/spotify/apollo/Response.java
```diff
@@ -40,10 +42,33 @@
...
    * The response headers.
+   * Deprecated in favor of {@link Response#headerEntries()}
    */
+  @Deprecated
   Map<String, String> headers();
 
   /**
+   * The response headers
+   */
+  Iterable<Map.Entry<String, String>> headerEntries();
```
pettermahlen: Why Iterable here? Since we're spending some effort on retaining order of headers, would it not make more sense to return a List?

apollo-api/src/main/java/com/spotify/apollo/Response.java
```diff
@@ -119,4 +144,5 @@
   static <T> Response<T> of(StatusType statusCode, T payload) {
     return ResponseImpl.create(statusCode, payload);
   }
+
```
pettermahlen: Probably remove this empty line

apollo-api/src/main/java/com/spotify/apollo/Headers.java
```diff
@@ -0,0 +1,82 @@
...
+      final Map.Entry<String, String> headerEntry = entries().get(i);
+      if (name.equalsIgnoreCase(headerEntry.getKey())) {
+        return Optional.ofNullable(headerEntry.getValue());
+      }
+    }
+
+    return Optional.empty();
+  }
+
+  public Map<String, String> asMap() {
```
martina-if: This is used a lot in apollo. Should I memoized it already in the constructor?

apollo-api/src/main/java/com/spotify/apollo/Response.java
martina-if: I was trying to be conservative and restrictive, conveying the message that this is only for iterating over all the headers. I want to prevent users from looking up keys from whatever object this method returns. Also, I didn't want to lock a specific implementation (we could change the internal implementation for a Map, etc).  But I see now that I chose the wrong interface and it should have been Iterator, but maybe it's too annoying, specially since it does not have size(). What do you think would be better: Iterator or List?

pettermahlen: Iterator is generally wrong, since they are one-use only. Iterable is better in that case.

I would still prefer List; it's still quite generic, and I feel like we have committed to retaining order. We could even specify that in the javadocs.

apollo-api/src/main/java/com/spotify/apollo/Headers.java
pettermahlen: Not sure; doing so would increase the memory footprint - you'd need some measurements to be sure, I guess.

apollo-api/src/main/java/com/spotify/apollo/Headers.java
```diff
@@ -0,0 +1,82 @@
...
+    ImmutableMap.Builder<String, String> headers = ImmutableMap.builder();
+    entries().forEach(headers::put);
+    return headers.build();
+  }
+
+  public abstract ImmutableList<Map.Entry<String, String>> entries();
+
+  private static void insertIfAbsent(List<Map.Entry<String, String>> headerList,
```
kouKatsumi: I'm confused here. Method name suggest that the value would not be overwritten (as it will be entered only if it's absent), but that's what the method does?

apollo-api/src/test/java/com/spotify/apollo/HeadersTest.java
```diff
@@ -0,0 +1,157 @@
...
+import static org.hamcrest.CoreMatchers.is;
+import static org.junit.Assert.assertThat;
+
+public class HeadersTest {
+
+  private static final Map<String, String> TEST_MAP_DUPLICATE_KEYS = createDuplicateKeysMap();
+  private static final Map<String, String> TEST_BIG_MAP = createBigMap();
+
+  private static Map<String, String> createDuplicateKeysMap() {
+    LinkedHashMap<String, String> testMap = new LinkedHashMap<>(3);
```
kouKatsumi: Maybe declare test map as`Map` instead of `LinkedHashMap` ?

apollo-api/src/test/java/com/spotify/apollo/ResponseTest.java
```diff
@@ -52,9 +55,39 @@ public void allowsAddingMultipleHeaders() {
...
         .withHeaders(headers);
 
-    assertEquals("application/protobuf", response.headers().get("Content-Type"));
-    assertEquals("123", response.headers().get("Content-Length"));
+    assertEquals(Optional.of("application/protobuf"), response.header("Content-Type"));
+    assertEquals(Optional.of("123"), response.header("Content-Length"));
+  }
+
+  @Test
+  public void shouldReturnHeaderCaseInsensitive() throws Exception {
```
kouKatsumi: Do we want to test it again as we already do in `HeadersTest.java`?

apollo-api/src/main/java/com/spotify/apollo/Headers.java
martina-if: Thanks! Fixed

apollo-api/src/main/java/com/spotify/apollo/Request.java
martina-if: Talked offline about this. I will leave the check with `Objects.requireNonNull`

apollo-api/src/main/java/com/spotify/apollo/Response.java
martina-if: Fair enough. Changed to `List`.

apollo-api/src/main/java/com/spotify/apollo/Response.java
martina-if: Done

apollo-api/src/main/java/com/spotify/apollo/Headers.java
martina-if: I'll leave the optimizations for later then.

apollo-api/src/main/java/com/spotify/apollo/Headers.java
martina-if: Right. I fixed the name to insertOrReplace(). Is this better?

apollo-api/src/test/java/com/spotify/apollo/HeadersTest.java
martina-if: Fixed

apollo-api/src/test/java/com/spotify/apollo/ResponseTest.java
martina-if: Yes, I think the test should be implementation agnostic. At least the most important parts.

PR #201, opened 2017-05-11
Add a minimum for reply rate in error ratio calculation

johanburatti:
This fixes an issue where the error-reply-ratio metric will show un-predictable behavior after sites or hosts have been drained without the hosts being restart afterwards.

This is a cherry-pick of this change from master. See #195 

@martina-if


kouKatsumi:
:+1:

PR #204, opened 2017-05-24
do not log 4xx errors as warnings

danielnorberg:
Application logs should not be filled with warnings because of bad user requests.

Log messages for 4xx requests should just go into the access logs.


mattnworb:
👍

martina-if:
I actually found those log lines very useful in the past. Would it make sense to leave them as `INFO: ignoring bad url...`?

danielnorberg:
@martina-if I guess it'd be nice to be able to somehow opt-in to logging or handling "bad" requests, but defaulting to logging bad requests seems undesirable.

martina-if:
@dano do we get this logs into the access logs?

danielnorberg:
@martina-if Yes, all requests are logged in the access log, including requests resulting in 4xx, 5xx etc.

martina-if:
:+1: Ok, then I think we have enough info to debug when this happens

PR #205, opened 2017-05-25
Mark only 500 codes as errors in metrics

gilles:
@martina-if @danielnorberg 

Slightly related to #204

### DIFF
--- a/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
+++ b/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
@@ -32,8 +32,7 @@ import java.util.function.Consumer;
 
 import okio.ByteString;
 
-import static com.spotify.apollo.StatusType.Family.INFORMATIONAL;
-import static com.spotify.apollo.StatusType.Family.SUCCESSFUL;
+import static com.spotify.apollo.StatusType.Family.SERVER_ERROR;
 import static java.util.Objects.requireNonNull;
 
 // Optional fields are fine; they enable the use of the 'ifPresent' idiom which is more readable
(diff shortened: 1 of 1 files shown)


martina-if:
I'm not sure about this. I think this will break people's dashboards silently. And since you can still group metrics by the status code I think it makes more sense to mark all of them as errors.

gilles:
The main issue is the ratio, which is unusable for us right now because we send a lot of 404 and only 3XX on some endpoints. I can introduce another meter and update the ratio to use the new one.

martina-if:
Hi @gilles, I would rather not change a metric for everyone for that specific case. Have you tried filtering those status codes (or families) from your graph with the extended syntax? I can help you set it up or you can ask the monitoring team how to do this.

Annother option could be to set up a custom metric for your service but I don't see the need for this additional metric right now.

OskarKjellin:
+1000 on this. I don't think people expect their error ratio to include non error status codes. They're most likely more broken now than they can get by merging this

mattnworb:
One option that could sidestep the backwards compatibility issue would be to add meters for each of the status families (Informational, Success, redirection, client error, server error)

martina-if:
@gilles I think that either the new ratio as you say or what matt suggested would be great. I'm leaning more towards the separate meters for each.

gilles:
@mattnworb you mean a new ratio gauge for each family? That would work. Do you want me to keep the aggregated existing one as well?

mattnworb:
I was suggesting adding new meters, so you could easily count 4xx responses differently from 5xx responses, but I don't really have an opinion on what happens with the existing ones. I suppose you would want to keep them for the backwards compatibility concerns mentioned earlier.

gimaker:
@gilles @martina-if I think this makes sense in general and I'd be happy to merge this for _master_, AKA the 2.x branch which has never yet been released (yep, it's confusing). But I agree with @martina-if that changing existing metrics is a no-no since that breaks "API". For the `1.x` branch we'd happily accept PRs that adds new, more usable metrics in coexist with the current metrics.

gilles:
fine with me

gimaker:
@gilles Do you want this merged (for master) or shall we close it?

gilles:
I think it should since this is the rate used to compute the ratio used for alerting. Since 2.0 is not released you can rethink the metrics but in case you can't get to it I think this is better than what we have now.

PR #206, opened 2017-06-13
Upgrade logback and slf4j

OskarKjellin:


martina-if:
Let me test this before merging, I don't trust it unless we have missinglink plugin or we try it out because there could be runtime dependencies missing

martina-if:
:+1:

PR #208, opened 2017-06-16
Read X-Calling-Service header in a case insensitive way

martina-if:
@kouKatsumi 

By using the method  `Request.header(String headerName)` instead of a lookup in the map we get case insensitive lookup of that header.


jpettersson:
:+1:

PR #209, opened 2017-06-18
Preserve case in com.spotify.apollo.Headers

dflemstr:
We own an application that gets requests directly from devices in the wild (that can't be upgraded).  These devices talk a great variety of different protocols (hm1, hm2, mercury multiget, http1, http2, ...) and some of these protocols are not case insensitive.

We eventually want to upgrade to the coming version of Apollo but with the current API it is not possible to respond with headers that have a certain case.

This change makes Apollo preserve the original case of all headers, with the trade-off that they are ordered alphabetically.  This was deemed okay since after a discussion and some investigation we concluded that the order of headers only matters when repeated headers are supported, which Apollo does not do.

Another benefit is that the backing map of the headers is immutable now, which prevents surprising behavior if it accidentally is modified.


martina-if:
@dflemstr This work was already done in apollo 1.4. This PR is against master which will be at some point apollo 2.0 but it's not yet released. Sorry for the confusion. Please try out v1.4.1 and let me know if you still have issues.

dflemstr:
@martina-if Will this not be needed for `2.x` as well?

martina-if:
@dflemstr For apollo 2.0 it was decided to use only lowercase headers. If this is a problem we can discuss it again and, possibly, port the current implementation to 2.0.

PR #211, opened 2017-06-27
Remove application-version from Metrics

dmichel1:
Merging this to the 1.x branch per the slack discussion. 

https://github.com/spotify/apollo/pull/189 was merged into master which is for the 2.x release.

\cc @martina-if


martina-if:
:+1:

PR #213, opened 2017-07-04
Bump to version 1.5.0

martina-if:
@kouKatsumi @jpettersson 

This bumps first the version to the 1.5 snapshot. Then I can do the release. Note that there is a test that protects version bumping without a explicit commit for bumping the version.


kouKatsumi:
:+1:

PR #215, opened 2017-08-31
Close all AutoCloseable

OskarKjellin:


apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
```diff
@@ -208,9 +203,15 @@ private InstanceImpl initInstance(
...
     for (Key<?> key : keysToLoad) {
       Object obj = injector.getInstance(key);
-      if (Closeable.class.isAssignableFrom(obj.getClass())) {
+      if (AutoCloseable.class.isAssignableFrom(obj.getClass())) {
         LOG.info("Managing lifecycle of {}", key.getTypeLiteral());
-        closer.register(Closeable.class.cast(obj));
+        closer.register(() -> {
+          try {
+            AutoCloseable.class.cast(obj).close();
+          } catch (final Exception ex) {
```
kouKatsumi: I'm a bit concerned with suppressing all `Exception`s. As `AutoClosable` javadoc says [https://docs.oracle.com/javase/8/docs/api/java/lang/AutoCloseable.html](here):

> Implementers of this interface are also strongly advised to not have the close method throw InterruptedException. This exception interacts with a thread's interrupted status, and runtime misbehavior is likely to occur if an InterruptedException is suppressed.

kouKatsumi:
LGTM, but maybe @gimaker could have a look as well?

gimaker:
@OskarKjellin Could you add a description of what problem this fixes to the PR? That would make it easier to review. Is it possible to add a test as well?

OskarKjellin:
The problem is that `FastForwardReporter` (https://github.com/spotify/semantic-metrics/blob/master/ffwd-reporter/src/main/java/com/spotify/metrics/ffwd/FastForwardReporter.java#L53) only implements `AutoCloseable`. Apollo seems to incorrectly think that it's managing the lifcecycle of the `FastFowardReporter`(https://github.com/spotify/apollo/blob/master/modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsModule.java#L80) but it doesn't, because of the bug this PR fixes.

apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
OskarKjellin: That documentation implies that this catch would never get an `InterruptedException` tho. Otherwise I agree

liljencrantz:
👍

apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
mattnworb: I think you could avoid this worry if the original `if closeable, closer.register((Closer)obj)` block was kept and this AutoCloseable block was added as a fallback.

OskarKjellin: I don't really see the reason to treat them differently? One simply extends the other and we can call the super

gimaker:
@OskarKjellin Thanks for the explanation! Makes sense. A regression test would be nice but :+1:

apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
mattnworb: my thinking was that in this case you are handling the exception itself, and in the previous state the closer is handling the exceptions and bubbling them up when `closer.close()` is called, but I am not sure if anyone or any part of Apollo is relying on that.

OskarKjellin: The `Closer` catches `Throwable` anything...

PR #217, opened 2017-10-10
Add config option for following redirects for the okhttpclient

reidwilbur:


reidwilbur:
@rouzwawi

mattnworb:
looks ok to me 👍

martina-if:
Thanks for the PR! It looks good :+1:. 

One important note @reidwilbur: we can merge this into master, which is the release 2.0 of apollo, and at the moment there are no plans or date for that release. If you want this feature in the next 1.x version of apollo you need to duplicate this PR against the branch `1.x`. (We need to merge both).

PR #219, opened 2017-10-12
Add config option for following redirects for the okhttpclient

reidwilbur:


reidwilbur:
@martina-if added the config option for okhttp client for the 1.x branch

martina-if:
:+1:

PR #220, opened 2017-10-19
upgrade to okhttp3

honnix:
also upgrade okio

okhttp 2.x is obsolete, and according to https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-300-rc1
there should be no risk to upgrade.

### DIFF
--- a/apollo-bom/pom.xml
+++ b/apollo-bom/pom.xml
@@ -97,14 +97,14 @@
             </dependency>
 
             <dependency>
-                <groupId>com.squareup.okhttp</groupId>
+                <groupId>com.squareup.okhttp3</groupId>
                 <artifactId>okhttp</artifactId>
-                <version>2.5.0</version>
+                <version>3.9.0</version>
             </dependency>
             <dependency>
                 <groupId>com.squareup.okio</groupId>
                 <artifactId>okio</artifactId>
-                <version>1.6.0</version>
+                <version>1.13.0</version>
             </dependency>
             <dependency>
                 <groupId>com.google.inject</groupId>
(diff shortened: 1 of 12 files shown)


honnix:
@martina-if Does this makes sense? Or you have tried but found some big issue?

honnix:
@mattnworb that makes perfect sense! not sure how this kind of change has been tracked usually in this project. I'm not a frequent committer here. :D

danielnorberg:
@mattnworb Projects that access okhttp2 directly _should_ declare the direct dependency on okhttp2 instead of relying on the indirect dependency through apollo. Projects that do not declare a direct dependency might get a compilation error when they upgrade apollo and they can then either add the direct dependency or (preferably) also switch to using okhttp3. This seems fair to me. Am I missing something? Wdyt?

mattnworb:
Right, they _should_ declare a direct dependency. My question was just if there is a way to document this change to help make it obvious at upgrade-time. 

I don't see a CHANGELOG in the repo but seems like the main maintainers have been able to document each release on [the releases page](https://github.com/spotify/apollo/releases) just fine without one, so perhaps it isn't that important.

honnix:
So ok to merge? @pettermahlen

danielnorberg:
@mattnworb Yeah, ideally maven would be a bit more helpful in these situations. To digress a little bit, I guess java modules would have prevented this situation to begin with. Anyway, I guess we should at least mention the okhttp change on the releases page.

martina-if:
@honnix Thank you for the PR! Master branch is for apollo 2.0 version. We don't have that release planned in the near future, did you mean to have it for the next 1.x release? In that case you should duplicate the PR against branch `1.x`. Sorry for the inconvenience, I know it's not obvious.

Regarding the compilation errors. I think that since they are compilation errors this is reasonable.

honnix:
This can be for both 2.0 and 1.x I think. How about getting it into 2.0 and back port to 1.x next release?

martina-if:
@honnix Yes, please, duplicate it to the other branch. I'll merge this one.

PR #221, opened 2017-10-19
do not duplicate authorization header

honnix:
instead of adding the header without check, reset it in favor of that
from incoming request.

Does this make sense? Do we have use case replying on multiple `Authorization` header?

### DIFF
--- a/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
+++ b/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
@@ -78,7 +78,11 @@ class HttpClient implements IncomingRequestAwareClient {
 
     apolloIncomingRequest
         .flatMap(req -> req.header(AUTHORIZATION_HEADER))
-        .ifPresent(header -> headersBuilder.set(AUTHORIZATION_HEADER, header));
+        .ifPresent(header -> {
+          if (headersBuilder.get(AUTHORIZATION_HEADER) == null) {
+            headersBuilder.add(AUTHORIZATION_HEADER, header);
+          }
+        });
 
     final Request request = new Request.Builder()
         .method(apolloRequest.method(), requestBody.orElse(null))
--- a/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
+++ b/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
@@ -73,12 +73,12 @@ class HttpClient implements IncomingRequestAwareClient {
       return RequestBody.create(contentType, payload);
     });
 
-    Headers.Builder headersBuilder = new Headers.Builder();
-    apolloRequest.headers().asMap().forEach((k, v) -> headersBuilder.add(k, v));
+    final Headers.Builder headersBuilder = new Headers.Builder();
+    apolloRequest.headers().asMap().forEach(headersBuilder::add);
 
     apolloIncomingRequest
         .flatMap(req -> req.header(AUTHORIZATION_HEADER))
-        .ifPresent(header -> headersBuilder.add(AUTHORIZATION_HEADER, header));
+        .ifPresent(header -> headersBuilder.set(AUTHORIZATION_HEADER, header));
 
     final Request request = new Request.Builder()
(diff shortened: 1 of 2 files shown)


honnix:
@martina-if PTAL

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
```diff
@@ -73,12 +73,12 @@ public static HttpClient createUnconfigured() {
...
 
-    Headers.Builder headersBuilder = new Headers.Builder();
-    apolloRequest.headers().asMap().forEach((k, v) -> headersBuilder.add(k, v));
+    final Headers.Builder headersBuilder = new Headers.Builder();
+    apolloRequest.headers().asMap().forEach(headersBuilder::add);
 
     apolloIncomingRequest
         .flatMap(req -> req.header(AUTHORIZATION_HEADER))
-        .ifPresent(header -> headersBuilder.add(AUTHORIZATION_HEADER, header));
+        .ifPresent(header -> headersBuilder.set(AUTHORIZATION_HEADER, header));
```
danielnorberg: Is it expected that the header from the incoming request overrides the header explicitly set on the new request? 🤔

honnix: Not sure how Apollo handles duplicated headers now. Will take a look.

honnix: It seems it joins multiple headers into a comma separated value: https://github.com/spotify/apollo/blob/master/modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java#L120

That would be interesting for header like `Authentication`.

danielnorberg:
When we introduced the automatic auth context propagation apollo was hermes-only so it was unlikely that credentials would be leaked somewhere by accident. Only spotify internal services talk hermes so presumably nobody would use the request context client to talk to external services.

The situation is a little different for http services. Automatic propagation of the `Authorization` header seems a bit more risky for http services as it is easy to use the request context client by mistake when talking to external services. I wonder if it would make sense to remove automatic `Authorization` header propagation entirely from the http client. 🤔

honnix:
Yeah, I would like to remove it. Not sure of the conversation you had last time with rouz. I can of course make another PR to remove it, but the consequence can be quite big/bad.

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
danielnorberg: Yeah, if we're going to send an `Authorization` header I think we want it to have only one value. My question was more about which value should take precedence. Personally I would expect the one explicitly set on the outgoing request to be used and not overridden by the value from the incoming request. Wdyt?

danielnorberg:
@pettermahlen @rouzwawi wdyt about removing automatic forwarding of `Authorization` header for http requests completely?

pettermahlen:
I'm out of touch and don't have a good opinion about that. @martina-if @jo-ri ?

martina-if:
This PR would be merged into master, which means that this would be released in the future on apollo 2.0. If you want to change that behavior for apollo 2.0 it makes sense. If you want that change in the next 1.x release I think it could be risky, and although I think not that many services use that forwarding I would prefer not changing it yet. 

@gimaker What do you think?

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClient.java
martina-if: Good catch, I also think that what the user sets should take precedence.

honnix: Agreed, that would make more sense.

gimaker:
makes sense to me :+1: as it is right now it seems un-risky - unless you explicitly set the `Authorization` header when making requests there is no change, IIUC.

I also agree it would make sense to _not_ automatically forward the `Authorization` header, but that's obviously a different, breaking change

honnix:
So, fine to merge? I can send another PR to 1.x branch afterwards.

PR #222, opened 2017-10-20
[Docs] Update middleware.md

gpiress:
I needed to decorate a RequestContext and it took me some time to figure out how to do this.

I ended up finding another project actually doing this to figure out how to do it, so I thought 
it might be good to have it more explicit in the docs.

### DIFF
--- a/apollo-api/docs/middleware.md
+++ b/apollo-api/docs/middleware.md
@@ -78,6 +78,29 @@ the handler type our middleware is defined to work on. In the end we apply the
 `Middleware::syncToAsync` middleware (defined in apollo-api) which turns the route handler into an
 `AsyncHandler<T>` which the framework can call._
 
+### Decorating a RequestContext
+
+If you need to decorate a request before passing invoking your inner handlers, you will need to
+create a new decorated `RequestContext`. This is how you can decorate a request with a custom
+header:
+
+```java
+private RequestContext decorateRequestContext(RequestContext originalRequestContext) {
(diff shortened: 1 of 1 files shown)


stefanalund:
👍

apollo-api/docs/middleware.md
```diff
@@ -78,6 +78,29 @@ the handler type our middleware is defined to work on. In the end we apply the
 `Middleware::syncToAsync` middleware (defined in apollo-api) which turns the route handler into an
 `AsyncHandler<T>` which the framework can call._
 
+### Decorating a RequestContext
+
+If you need to decorate a request before passing invoking your inner handlers, you will need to
```
martina-if: "...before ~passing~ invoking your inner handlers..." ?

apollo-api/docs/middleware.md
```diff
@@ -78,6 +78,29 @@ the handler type our middleware is defined to work on. In the end we apply the
...
+create a new decorated `RequestContext`. This is how you can decorate a request with a custom
+header:
+
+```java
+private RequestContext decorateRequestContext(RequestContext originalRequestContext) {
+  Map<String, String> decoratedRequestHeaders = originalRequestContext.request().headers();
+  decoratedRequestHeaders.put("Custom-Header", "Custom value");
```
martina-if: The `headers()` method is deprecated. Please use `headerEntries()` instead.

gpiress: @martina-if where can I find this? Does it return a `Map<String, String>` as the old `headers()` method?

PR #224, opened 2017-10-24
upgrade to okhttp3

honnix:
also upgrade okio

okhttp 2.x is obsolete, and according to https://github.com/square/okhttp/blob/master/CHANGELOG.md#version-300-rc1
there should be no risk to upgrade.
  
backport of #220

### DIFF
--- a/apollo-bom/pom.xml
+++ b/apollo-bom/pom.xml
@@ -92,14 +92,14 @@
             </dependency>
 
             <dependency>
-                <groupId>com.squareup.okhttp</groupId>
+                <groupId>com.squareup.okhttp3</groupId>
                 <artifactId>okhttp</artifactId>
-                <version>2.5.0</version>
+                <version>3.9.0</version>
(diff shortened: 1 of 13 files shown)


honnix:
This is probably a breaking change though. okhttp3 default uses lower case for header key?

danielnorberg:
@honnix header keys should be treated in a case-insensitive manner though. Are we not doing that?

honnix:
Not sure of that and hard to presume.

pettermahlen:
#199

honnix:
@pettermahlen that feels safe already in Apollo world. But it seems okhttp3 lowers down the case already when building headers, which means things going out of Apollo (and potentially to other system) are different now. If everyone behaves according to RFC, that should be ok though.

honnix:
#209  :)

honnix:
this however doesn't pass `japicmp`

PR #232, opened 2017-12-08
New error ratio metrics (gauges) for 4xx and 5xx service responses

viktorjonsson:


modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
```diff
@@ -92,9 +101,15 @@ public void response(Response<ByteString> response) {
     timerContext.ifPresent(Timer.Context::stop);
 
     StatusType.Family family = response.status().family();
-    if (family == SERVER_ERROR) {
+    if (family != INFORMATIONAL && family != SUCCESSFUL) {
```
gilles: redirects should not count as errors either

gilles:
someone beats me in making the 4xx and 5xx ratios :) thanks!

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
viktorjonsson: The reason why I added two new metrics was that I didn't want to touch the error-ratio one (as was discussed in this PR https://github.com/spotify/apollo/pull/205). I'd prefer not to touch that metric.

mattnworb:
If you want to merge this into the 1.x branch in #233, shouldn’t you merge it into master as well for whatever comes after 1.x?

PR #233, opened 2017-12-11
New error ratio metrics (gauges) for 4xx and 5xx service responses

viktorjonsson:
Updated docs for new response error metrics

### DIFF
--- a/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
+++ b/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
@@ -33,7 +33,9 @@ import java.util.function.Consumer;
 import okio.ByteString;
 
 import static com.spotify.apollo.StatusType.Family.INFORMATIONAL;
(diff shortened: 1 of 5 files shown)


modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticRequestMetrics.java
```diff
@@ -33,7 +33,10 @@
 import okio.ByteString;
 
 import static com.spotify.apollo.StatusType.Family.INFORMATIONAL;
+import static com.spotify.apollo.StatusType.Family.SERVER_ERROR;
 import static com.spotify.apollo.StatusType.Family.SUCCESSFUL;
+import static com.spotify.apollo.StatusType.Family.CLIENT_ERROR;
+import static com.spotify.apollo.StatusType.Family.REDIRECTION;
```
gimaker: Looks like an unused import?

viktorjonsson: thanks! I removed it

PR #239, opened 2018-01-17
For error ratios below 1e-15, report them as zero

gimaker:
Small but not-quite zero error ratios can, at a glance, look alarming
in graphs. An increase from e.g. 1e-316 to 1e-300 looks like a big scary
increase, when in practice the error ratio is still basically nil.

Because the error ratios are calculated via EWMA-computed rates they
can exhibit small fluctuations in hard-to-understand ways.


gimaker:
Same patch for 1.x branch: #240

PR #241, opened 2018-01-20
1.x: build with jdk10 and fix compilation errors

danielnorberg:


danielnorberg:
```
[ERROR] /Users/dano/projects/apollo/apollo-entity/src/main/java/com/spotify/apollo/entity/CodecEntityMiddleware.java:[155,31] incompatible types: bad return type in lambda expression
[ERROR]     com.spotify.apollo.Response<R> cannot be converted to capture#3 of ? extends com.spotify.apollo.Response<R>
[ERROR] /Users/dano/projects/apollo/apollo-entity/src/main/java/com/spotify/apollo/entity/CodecEntityMiddleware.java:[164,46] incompatible types: bad return type in lambda expression
[ERROR]     no instance(s) of type variable(s) U exist so that java.util.concurrent.CompletableFuture<U> conforms to capture#6 of ? extends java.util.concurrent.CompletionStage<com.spotify.apollo.Response<R>>
```

🤔

danielnorberg:
@gimaker PTAL

jhaals:
This looks good to me. Scariest thing is the guava bump but I guess most people are managing Guava themselves anyway. thoughts @tommyulfsparre  or @fredah?

PR #244, opened 2018-01-26
Attempt to read payload even when contentLength = -1

### LINKED ISSUE #243: Requests with "Transfer-Encoding: chunked" wrongly conclude that no payload is being sent
When sending a request with the header [Transfer-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding): chunked, Apollo is concluding that no payload is being sent due to the fact that getting the content length of such request may return -1 (unknown length). Such header indicates that the payload is being sent as a series of chunks.

I think that Apollo should still try to read the streamed payload in such a case before concluding that it is empty. I am wrapping up a fix for this which I expect to submit later today.

excentris:
When sending a request with the header Transfer-Encoding: chunked,
Apollo is concluding that no payload is being sent due to the fact that
getting the content length of such request may return -1 (unknown length).
Such header indicates that the payload is being sent as a series of chunks.

This change tries to read the streamed payload in such a case before
concluding that it is empty.

fixes #243

### DIFF
--- a/modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java
+++ b/modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java
@@ -21,14 +21,18 @@ package com.spotify.apollo.http.server;
 
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableMap;
-
+import com.google.common.io.ByteStreams;
 import com.spotify.apollo.Request;
 import com.spotify.apollo.RequestMetadata;
 import com.spotify.apollo.request.RequestHandler;
 import com.spotify.apollo.request.RequestMetadataImpl;
-
+import okio.ByteString;
 import org.eclipse.jetty.server.handler.AbstractHandler;
 
+import javax.servlet.AsyncContext;
+import javax.servlet.ServletException;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
 import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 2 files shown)


gimaker:
Thanks for the PR! And apologies for the slow response time.

excentris:
@gimaker no problem!

gimaker:
@rouzwawi @pettermahlen @fredah @jhaals @tommyulfsparre @martina-if

modules/jetty-http-server/src/main/java/com/spotify/apollo/http/server/ApolloRequestHandler.java
```diff
@@ -121,24 +112,33 @@ Request asApolloRequest(HttpServletRequest req) throws IOException {
...
-  private ByteString readPayload(HttpServletRequest req, int contentLength) throws IOException {
+  private Optional<ByteString> readPayload(HttpServletRequest req, int contentLength) throws IOException {
     final InputStream input = new BufferedInputStream(req.getInputStream());
-    return ByteString.read(input, contentLength);
+    if (contentLength < 0) {
+      // contentLength = -1 may be returned when using Transfer-Encoding: chunked
+      // (RFC 7230, section 3.3.1: Transfer-Encoding) even though data is being transferred in a series of chunks.
+      // In that case try read this data before concluding that there is no payload.
+      final ByteString byteString = ByteString.of(ByteStreams.toByteArray(input));
+      return byteString.size() == 0 ? Optional.empty() : Optional.of(byteString);
```
pettermahlen: On reading this PR, I notice that apollo treats content-length = 0 as 'payload present but empty'. It looks to me like this code means that the case of (chunked + content-length = -1 + content actually has zero length) leads to different behaviour. Is there a way of ensuring that the behaviour would be identical in that case? There doesn't seem to be any unit tests guarding the old behaviour either. And it's unlikely to be a very important distinction I guess, so I'm fine with deciding to not fix that.

excentris: Hi! I currently don't have much time to look into this. I can however say that we've been running a patched version (with the code on this PR) of Apollo for a few months now and have had no problem. I think I could take a further look into this at some point when I'm less busy, but I agree it would be fine to merge this as is.

PR #247, opened 2018-02-09
Allow entity codecs access to the RequestContext

rouzwawi:
It's usually the case that services want to support multiple payload formats for routes. Today, the `EntityMiddleware` utility is limited to codecs for a single format, and it's impossible to implement any type of dynamic behaviour (e.g. based on `Content-Type` or `Accept` headers) on the `EntityCodec` api. The way out is to duplicate lots of the `CodecEntityMiddleware` implementation just do get more dynamic serialization.

This PR addresses this by introducing a new `Codec` interface that receives the `RequestContext` and can thus do more dynamic serialization. The old `EntityCodec` is marked as `@Deprecated`, but all previous functionality is left intact.


apollo-entity/src/main/java/com/spotify/apollo/entity/CodecEntityMiddleware.java
```diff
@@ -19,79 +19,73 @@
...
 
   @Override
   public <R> Middleware<SyncHandler<R>, SyncHandler<Response<ByteString>>>
   serializerDirect(Class<? extends R> responseEntityClass) {
-    return inner -> inner
+    return inner -> rc -> inner
         .map(Response::forPayload)
-        .map(serialize(responseEntityClass));
+        .map(serialize(rc, responseEntityClass))
+        .invoke(rc);
```
pettermahlen: Wow. That's kind of mind-boggling. Is there a more obvious way to write this code, maybe?

rouzwawi: I know it's a bit dense. I tried expanding it a bit but I have a hard time finding meaningful names for the intermediate handlers. Here's an example:

```java
@Override
public <R> Middleware<SyncHandler<R>, SyncHandler<Response<ByteString>>>
    serializerDirect(Class<? extends R> responseEntityClass) {
  return inner -> {
    final SyncHandler<Response<ByteString>> wrappedHandler = rc -> {
      final SyncHandler<Response<ByteString>> serializingHandler = inner
          .map(Response::forPayload)
          .map(serialize(rc, responseEntityClass));

      return serializingHandler.invoke(rc);
    };
    return wrappedHandler;
  };
}
```

or this one that is slightly less expanded:

```java
@Override
public <R> Middleware<SyncHandler<R>, SyncHandler<Response<ByteString>>>
    serializerDirect(Class<? extends R> responseEntityClass) {
  return inner -> rc -> {
    final SyncHandler<Response<ByteString>> serializingHandler = inner
        .map(Response::forPayload)
        .map(serialize(rc, responseEntityClass));

    return serializingHandler.invoke(rc);
  };
}
```

pettermahlen: I like the second option, that is easier for me to read.

rouzwawi: Refactored 👍

PR #248, opened 2018-02-09
Allow entity codecs access to the RequestContext (1.x)

rouzwawi:
Backport of #247 for the `1.x` branch

### DIFF
--- a/apollo-entity/src/main/java/com/spotify/apollo/entity/CodecEntityMiddleware.java
+++ b/apollo-entity/src/main/java/com/spotify/apollo/entity/CodecEntityMiddleware.java
@@ -19,32 +19,30 @@
  */
 package com.spotify.apollo.entity;
 
+import static java.util.concurrent.CompletableFuture.completedFuture;
+
 import com.spotify.apollo.RequestContext;
 import com.spotify.apollo.Response;
 import com.spotify.apollo.Status;
 import com.spotify.apollo.route.AsyncHandler;
 import com.spotify.apollo.route.Middleware;
 import com.spotify.apollo.route.SyncHandler;
-
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
 import java.io.IOException;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
-
 import javaslang.control.Either;
+import javax.annotation.Nullable;
 import okio.ByteString;
-
-import static java.util.concurrent.CompletableFuture.completedFuture;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A factory for creating middlewares that can be used to create routes that deal directly
  * with an api entity.
  *
- * A {@link EntityCodec} is used to define how to go between a {@link ByteString} and the
- * entity type used in your route handlers.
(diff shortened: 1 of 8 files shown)


apollo-entity/src/main/java/com/spotify/apollo/entity/CodecEntityMiddleware.java
```diff
@@ -19,79 +19,87 @@
...
 
-  CodecEntityMiddleware(EntityCodec codec) {
-    this(codec, codec.defaultContentType());
+  CodecEntityMiddleware(Codec codec) {
+    this.codec = Objects.requireNonNull(codec);
+    this.contentType = null;
   }
 
-  CodecEntityMiddleware(EntityCodec codec, String contentType) {
+  @Deprecated
```
pettermahlen: Maybe add a pointer to a description for the rationale for deprecation, and what should be used instead.

apollo-entity/src/main/java/com/spotify/apollo/entity/EntityCodec.java
```diff
@@ -19,16 +19,15 @@
...
-import com.spotify.apollo.Exploratory;
-
 import java.io.IOException;
-
 import okio.ByteString;
 
 /**
  * Interface for serializing and de-serializing entity types.
+ *
+ * @deprecated Use {@link Codec}
```
pettermahlen: Add "as that allows for content negotiation" or something similar?

apollo-entity/src/test/java/com/spotify/apollo/entity/CodecTest.java
```diff
@@ -0,0 +1,182 @@
...
+import com.spotify.apollo.Status;
+import com.spotify.apollo.entity.EntityCodecsTest.Entity;
+import com.spotify.apollo.route.Middleware;
+import com.spotify.apollo.route.Route;
+import com.spotify.apollo.test.ServiceHelper;
+import io.norberg.automatter.jackson.AutoMatterModule;
+import okio.ByteString;
+import org.junit.Test;
+
+public class CodecTest {
```
pettermahlen: This is more like `EntityMiddlewareTest`, I think? Codec tests would validate code behaviours, such as whether they perform content negotiation correctly, etc.

rouzwawi: Hmm, there's no actual implementation of content negotiation in the codebase right now. I've done a very simple assertion that the `RequestContext` is being passed into the `Codec` though. I assume that should be enough to allow any code to make decisions on it's own based on that.

pettermahlen: Absolutely, I just meant I don't feel like this class is actually testing codecs, it's testing `EntityMiddleware` ?

rouzwawi: That's true

rouzwawi: Renamed test class

apollo-entity/src/main/java/com/spotify/apollo/entity/EntityCodec.java
rouzwawi: Done

PR #251, opened 2018-04-04
Add ability to track number of requests hitting a duration goal by

jsferrei:
This code adds the ability for Apollo users to configure latency goals for their endpoints, and tracks how many requests meet these goals in a meter.

### DIFF
--- a/modules/metrics/README.md
+++ b/modules/metrics/README.md
@@ -118,7 +118,8 @@ A Meter, tagged with:
 | what        | "endpoint-request-duration-threshold-rate"     | Enable/disable with endpoint-duration-goal config           |
 | threshold   | *                 | The request duration goal, in milliseconds                   |
 
-If a duration goal is set in configuration (`endpoint-duration-goal`), this meter will be marked when a request meets its goal.
+This meter will only be created if a duration goal is set in the configuration (`endpoint-duration-goal`). The meter will be marked when a request meets its goal.
+
 
 ## [ffwd](https://github.com/spotify/ffwd) reporter
 
@@ -220,12 +221,15 @@ key | type | required | note
 --- | ---- | -------- | ----
 `endpoint-duration-goal.{endpoint}.{method}` | int | required | Duration threshold in milliseconds for an endpoint/method combination
 
+**Note** If an endpoint has a parameter it needs to be included. For example a route defined as `GET:/example/<name>` would have
+be configured as `/v1/example-b/<name>.GET`
+
 #### Example
 
 ```
 endpoint-duration-goal = {
-  /v1/test-a.GET = 200 # meter marked every time this endpoint resolves in under 200 ms
-  /v1/test-b.GET = 100 # meter marked every time this endpoint resolves in under 100 ms
+  /v1/example-a.GET        = 200 # meter marked every time this endpoint resolves in under 200 ms
+  /v1/example-b/<name>.GET = 100 # meter marked every time this endpoint with a named parameter resolves in under 100 ms
 }
 ```
 
--- a/modules/metrics/README.md
+++ b/modules/metrics/README.md
@@ -98,7 +98,7 @@ the last 5 minutes, etc. The metrics do not include dropped requests.
 
 ### Dropped Requests
 
-A Meter, tagged with: 
+A Meter, tagged with:
 
 | tag         | value                      | comment                                              |
 |-------------|----------------------------|------------------------------------------------------|
@@ -106,9 +106,19 @@ A Meter, tagged with:
 | unit        | "request"                  |                                                      |
 
 Requests are dropped when they expire: if they have a time-to-live and are older than
-that TTL, Apollo will not try to respond. Apollo may also drop requests if it is 
+that TTL, Apollo will not try to respond. Apollo may also drop requests if it is
 overloaded and cannot respond to all incoming requests.
 
+### Endpoint Duration Goal
+
+A Meter, tagged with:
+
+| tag         | value                      | comment                                              |
+|-------------|----------------------------|------------------------------------------------------|
+| what        | "endpoint-request-duration-threshold-rate"     | Enable/disable with endpoint-duration-goal config           |
+| threshold   | *                 | The request duration goal, in milliseconds                   |
+
+If a duration goal is set in configuration (`endpoint-duration-goal`), this meter will be marked when a request meets its goal.
 
 ## [ffwd](https://github.com/spotify/ffwd) reporter
 
@@ -122,15 +132,16 @@ key | type | required | note
 `metrics.server` | string list | optional | list of [`What`](src/main/java/com/spotify/apollo/metrics/semantic/What.java) names to enable; defaults to [ENDPOINT_REQUEST_RATE, ENDPOINT_REQUEST_DURATION, DROPPED_REQUEST_RATE, ERROR_RATIO]
 `metrics.precreate-codes` | int list | optional | list of status codes to precreate request-rate meters for, default empty
 `ffwd.type` | string | optional | indicates which type of ffwd reporter to use. Available types are `agent` and `http`. see below for details. default is `agent`.
+`endpoint-duration-goal`  | int map  | optional |  sets request duration thresholds in milliseconds to track how many requests meet a duration objective
 
-You may not want to enable all the metrics Apollo can create, since some of them can be expensive 
+You may not want to enable all the metrics Apollo can create, since some of them can be expensive
 (in particular on the alerting and graphing side), hence the ability to configure which
 metrics to emit via `metrics.server`.
 
 The `metrics.precreate-codes` setting is there since the request-rate meters
 are lazily created. Apollo doesn't emit any metrics for status codes it hasn't
 seen. This can lead to strange effects when, for instance, an error shows up
-for the first time after a restart. Pre-creating meters for status codes you 
+for the first time after a restart. Pre-creating meters for status codes you
 want to alert on makes it less likely to get false positives, since Apollo
 will then emit a '0' value until the first time a certain status code shows up.
 
@@ -203,6 +214,33 @@ ffwd.discovery = {
 }
 ```
 
+### endpoint-duration-goal
+
+key | type | required | note
+--- | ---- | -------- | ----
+`endpoint-duration-goal.{endpoint}.{method}` | int | required | Duration threshold in milliseconds for an endpoint/method combination
+
+#### Example
+
+```
+endpoint-duration-goal = {
+  /v1/test-a.GET = 200 # meter marked every time this endpoint resolves in under 200 ms
+  /v1/test-b.GET = 100 # meter marked every time this endpoint resolves in under 100 ms
+}
+```
+
+key | type | required | note
+--- | ---- | -------- | ----
+`endpoint-duration-goal.all-endpoints` | int | required | Global duration threshold in milliseconds, will be overridden by endpoint specific goals
+
+#### Example
+
+```
+endpoint-duration-goal = {
+  all-endpoints = 500 # unless individually overridden, a goal of 500ms for all endpoints
+}
+```
+
 ## Custom Metrics
 
 To set up custom metrics on the application level you'll need to get hold of the
@@ -229,4 +267,3 @@ for some description of how to do request handling decorations.
 For client-side metrics, wrap the ```Client``` you get from the ```RequestContext``` in a similar
 way to how ```DecoratingClient``` is implemented. In your wrapper, ensure that the right metrics
 are tracked.
-
--- a/modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsModule.java
+++ b/modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsModule.java
@@ -127,12 +127,12 @@ public class MetricsModule extends AbstractApolloModule {
   @Provides
   @Singleton
   public MetricsFactory apolloMetrics(
-      SemanticMetricRegistry metricRegistry, MetricsConfig metricsConfig
-  ) {
+      SemanticMetricRegistry metricRegistry, MetricsConfig metricsConfig) {
     return new SemanticMetricsFactory(
         metricRegistry,
         what -> metricsConfig.serverMetrics().contains(what),
-        metricsConfig.precreateCodes()
+        metricsConfig.precreateCodes(),
+        metricsConfig.durationThresholdConfig()
     );
   }
 
--- a/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/DurationThresholdConfig.java
+++ b/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/DurationThresholdConfig.java
@@ -57,14 +57,10 @@ public class DurationThresholdConfig {
   }
 
   private static Optional<Integer> parseAllEndpoints(final Config config) {
-    final Optional<Integer> allEndpoints;
-    final Config durationGoalConfig = config.getObject("endpoint-duration-goal").toConfig();
-    if (durationGoalConfig.hasPath("all-endpoints")) {
-      allEndpoints = Optional.of(durationGoalConfig.getInt("all-endpoints"));
-    } else {
-      allEndpoints = Optional.empty();
+    if (config.hasPath("endpoint-duration-goal.all-endpoints")) {
+      return Optional.of(config.getInt("endpoint-duration-goal.all-endpoints"));
     }
-    return allEndpoints;
+    return Optional.empty();
   }
 
   public static DurationThresholdConfig parseConfig(final Config config) {
--- a/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/DurationThresholdConfig.java
+++ b/modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/DurationThresholdConfig.java
@@ -1,0 +1,88 @@
+/*
+ * -\-\-
+ * Spotify Apollo Metrics Module
+ * --
+ * Copyright (C) 2013 - 2016 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+package com.spotify.apollo.metrics.semantic;
+
+import com.google.common.collect.ImmutableMap;
+import com.typesafe.config.Config;
+import com.typesafe.config.ConfigValue;
+import java.util.Collection;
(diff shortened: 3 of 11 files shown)


modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
```diff
@@ -191,6 +197,31 @@ private CachedMeters metersForEndpoint(String endpoint) {
...
+   */
+  private Optional<LatencySloTracker> requestLatencySloTracker(MetricId id, String endpoint) {
+    final String[] endpointComponents = endpoint.split(":");
+    final String key = String.format("latency-slo-metrics.%s", endpointComponents[1],
+        endpointComponents[0]);
+    final String allKey = "latency-slo-metrics.all-endpoints";
+    final boolean hasEndpointKey = config.hasPath(key);
+    final boolean hasAllKey = config.hasPath(allKey);
+    if (hasEndpointKey || hasAllKey) {
+      final int threshold = hasEndpointKey ? config.getInt(key) : config.getInt(allKey);
```
frecon: @jsferrei Does this mean it would be possible to add something like this to the conf?
 
```
latency-slo-metrics.endpoint1: 50
latency-slo-metrics.endpoint2: 150
latency-slo-metrics.all-endpoints: 200
```

frecon:
Perhaps add some usage documentation.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
jsferrei: @frecon Yes

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
```diff
@@ -191,6 +197,31 @@ private CachedMeters metersForEndpoint(String endpoint) {
...
+   *
+   * @param id
+   * @param endpoint
+   * @return
+   */
+  private Optional<LatencySloTracker> requestLatencySloTracker(MetricId id, String endpoint) {
+    final String[] endpointComponents = endpoint.split(":");
+    final String key = String.format("latency-slo-metrics.%s", endpointComponents[1],
+        endpointComponents[0]);
+    final String allKey = "latency-slo-metrics.all-endpoints";
```
frecon: Looks like there's not tests covering `latency-slo-metrics.all-endpoints` and `latency-slo-metrics.SOME_ENDPOINT`.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/LatencySloTracker.java
```diff
@@ -0,0 +1,51 @@
...
+import com.spotify.metrics.core.MetricId;
+import com.spotify.metrics.core.SemanticMetricRegistry;
+
+public class LatencySloTracker {
+
+  private final Meter latencyMeter;
+  private final int threshold;
+
+  public LatencySloTracker(MetricId id, SemanticMetricRegistry
+      metricRegistry, Integer threshold) {
```
mattnworb: `threshold` should be an `int` if you are going to unbox it to store it in `private final int threshold`

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
mattnworb: maybe the threshold should be included in one of the tags of the metric? I'm thinking about what happens if I change the threshold in the configuration, and then the reported metrics look very different

jsferrei: @mattnworb good call

jsferrei:
@fredah / @gimaker would either of you mind reviewing this?

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
```diff
@@ -191,6 +197,31 @@ private CachedMeters metersForEndpoint(String endpoint) {
...
 
+  /**
+   * Checks for latency-slo-metrics configuration options and sets up
+   * metrics that track how many requests meet a certain latency goal.
+   *
+   * @param id
+   * @param endpoint
+   * @return
+   */
+  private Optional<LatencySloTracker> requestLatencySloTracker(MetricId id, String endpoint) {
```
dmichel1: is the endpoint just the name of the route or does it include the method too? would be nice to include that in the javadoc.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/MetricsConfig.java
```diff
@@ -42,6 +43,7 @@
       EnumSet.of(
           ENDPOINT_REQUEST_RATE,
           ENDPOINT_REQUEST_DURATION,
+          ENDPOINT_REQUEST_LATENCY_SLO_RATE,
```
dmichel1: nitpick - should we just call everything a duration to remain consistent with the other metric names here? `ENDPOINT_REQUEST_DURATION_THRESHOLD_RATE`

Also I think this is very useful outside of the SLO use case since these duration metrics can actually be aggregated across instances. What do you think of a mass rename from `latency-slo-metrics` to `duration-threshold-metrics` or something similar?

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
dmichel1: @jsferrei that config example is missing the method (?) what happens if you leave the method off.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/MetricsConfig.java
jsferrei: 👍

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
jsferrei: It has to include the method

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
jsferrei: Method is required and will always be passed in as part of `endpoint`, I believe. The above example should include method.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
```diff
@@ -191,6 +197,31 @@ private CachedMeters metersForEndpoint(String endpoint) {
...
+   *
+   * @param id
+   * @param endpoint formatted as METHOD:URI
+   * @return
+   */
+  private Optional<DurationThresholdTracker> requestDurationThresholdTracker(MetricId id, String endpoint) {
+    final String[] endpointComponents = endpoint.split(":");
+    final String key = String.format("duration-threshold-metrics.%s.%s", endpointComponents[1],
+        endpointComponents[0]);
+    final String allKey = "duration-threshold-metrics.all-endpoints";
```
tommyulfsparre: `duration-threshold-metrics.all-endpoints` isn't documented.

modules/metrics/README.md
```diff
@@ -203,6 +204,22 @@ ffwd.discovery = {
 }
 ```
 
+### duration-threshold-metrics
+
+key | type | required | note
+--- | ---- | -------- | ----
+`duration-threshold-metrics.{endpoint}.{method}` | int | required | Duration threshold in milliseconds for an endpoint/method combination
```
tommyulfsparre: I can't tell by looking at this name what it is supposed to convey, also having `metrics` in the name is redundant.

Maybe a better name would be endpoint-duration-goal/endpoint-duration-slo ?

modules/metrics/README.md
```diff
@@ -203,6 +204,22 @@ ffwd.discovery = {
...
+
+#### Example
+
+```
+duration-threshold-metrics = {
+  /v1/test-a.GET = 200 # meter marked every time this endpoint resolves in under 200 ms
+  /v1/test-b.GET = 100 # meter marked every time this endpoint resolves in under 100 ms
```
tommyulfsparre: Can you add an example on how the final metrics would look like under the `Metrics` section?

modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsModule.java
```diff
@@ -127,12 +127,13 @@ public SemanticMetricRegistry semanticMetricRegistry() {
...
   public MetricsFactory apolloMetrics(
-      SemanticMetricRegistry metricRegistry, MetricsConfig metricsConfig
+      SemanticMetricRegistry metricRegistry, MetricsConfig metricsConfig, Config config
   ) {
     return new SemanticMetricsFactory(
         metricRegistry,
         what -> metricsConfig.serverMetrics().contains(what),
-        metricsConfig.precreateCodes()
+        metricsConfig.precreateCodes(),
+        config
```
tommyulfsparre: Why did you choose to pass in the entire `Config` object instead of adding the methods you need to the `MetricsConfig`?

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
```diff
@@ -191,6 +197,31 @@ private CachedMeters metersForEndpoint(String endpoint) {
...
+  /**
+   * Checks for duration-threshold-metrics configuration options and sets up
+   * metrics that track how many requests meet a certain duration threshold goal.
+   *
+   * @param id
+   * @param endpoint formatted as METHOD:URI
+   * @return
+   */
+  private Optional<DurationThresholdTracker> requestDurationThresholdTracker(MetricId id, String endpoint) {
+    final String[] endpointComponents = endpoint.split(":");
```
tommyulfsparre: What happens when the endpoint doesn't contains `:` ?

jsferrei: @tommyulfsparre I looked and it seems that endpoint always contains this, as it is passed from https://github.com/spotify/apollo/blob/master/modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsCollectingEndpointRunnableFactoryDecorator.java#L46 which calls `EndpointInfo.getName()` which always concats method and route: https://github.com/spotify/apollo/blob/725ba23c2dcc2ee0aecbf669e2166d76c4ce39cc/apollo-api-impl/src/main/java/com/spotify/apollo/dispatch/EndpointInfo.java#L60

It may be a better idea to refactor to pass down method and route separately. Happy to do that, if you agree.

tommyulfsparre: Nah, It fine as is. The other public method didn't have any documentation either and that code is unlikely to change.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/DurationThresholdConfig.java
```diff
@@ -0,0 +1,83 @@
...
+      Integer>> goals;
+
+  private DurationThresholdConfig(final Optional<Integer> allEndpoints,
+                                  final Map<String, Map<String,
+                                      Integer>> goals) {
+    this.allEndpoints = allEndpoints;
+    this.goals = goals;
+  }
+
+  public Optional<Integer> getDurationThresholdForEndpoint(final String method,
```
tommyulfsparre: This might be cleaner if you flatten the nested map into an `Map<String, Optional<Integer>>` and use the endpoint as the key in the form you expect: `METHOD:URI`.

If you treat allEndpoints as the default value this could be simplified to something like this:
```
return thresholds.getOrDefault(endpoint, allEndpoints);
```

Does that make sense?

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/DurationThresholdConfig.java
```diff
@@ -0,0 +1,83 @@
...
+  public Optional<Integer> getDurationThresholdForEndpoint(final String endpointMethod) {
+    final String[] endpointMethodSplit = endpointMethod.split(":");
+    return getDurationThresholdForEndpoint(endpointMethodSplit[0], endpointMethodSplit[1]);
+  }
+
+  public static DurationThresholdConfig parseConfig(final Config config) {
+    final Optional<Integer> allEndpoints;
+    final Map<String, Map<String, Integer>> goals;
+    if (config.hasPath("endpoint-duration-goal")) {
+      final ConfigObject configObject = config.getObject("endpoint-duration-goal");
```
tommyulfsparre: I believe you can write this as:
```
      config.getObject("endpoint-duration-goal")
          .withoutKey("all-endpoints")
          .unwrapped()
          .entrySet()
          .stream()
          .collect(
              Collectors.toMap(
                  entry -> doSomething(entry.getKey()),
                  entry -> doSomething((Map<String, Integer>) entry.getValue())));
```

which is also where you could flatten the map.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/SemanticServiceMetrics.java
```diff
@@ -187,10 +192,27 @@ private CachedMeters metersForEndpoint(String endpoint) {
...
 
+  /**
+   * Checks for endpoint-duration-goal configuration options and sets up
+   * metrics that track how many requests meet a certain duration threshold goal.
+   *
+   * @param endpoint formatted as METHOD:URI
+   */
+  private Optional<DurationThresholdTracker> requestDurationThresholdTracker(MetricId id,
+                                                                             String endpoint) {
+    final Optional<Integer> endpointThreshold =
```
tommyulfsparre: This can be written as:
```
    return durationThresholdConfig.getDurationThresholdForEndpoint(endpoint)
        .map(threshold -> Optional.of(new DurationThresholdTracker(id, metricRegistry, threshold)))
        .orElse(Optional.empty());
```

jsferrei:
@tommyulfsparre Thanks again for the feedback. This is much simpler now that we are pre-flattening the config. Wasn't quite able to use the snippet you provided, but something close.

modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsModule.java
```diff
@@ -127,12 +127,13 @@ public SemanticMetricRegistry semanticMetricRegistry() {
   @Provides
   @Singleton
   public MetricsFactory apolloMetrics(
-      SemanticMetricRegistry metricRegistry, MetricsConfig metricsConfig
+      SemanticMetricRegistry metricRegistry, MetricsConfig metricsConfig, Config config
```
tommyulfsparre: Config isn't used anymore.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/DurationThresholdConfig.java
```diff
@@ -0,0 +1,89 @@
...
+                                  final Map<String, Optional<Integer>> goals) {
+    this.allEndpoints = allEndpoints;
+    this.goals = goals;
+  }
+
+  public Optional<Integer> getDurationThresholdForEndpoint(final String endpointMethod) {
+    return goals.getOrDefault(endpointMethod, allEndpoints);
+  }
+
+  public static Set<Map.Entry<String, Optional<Integer>>> flattenNestedConfig(Map.Entry<String,
```
tommyulfsparre: This linebreak looks weird, can you move the parameter to its own line:
```
public static Set<Map.Entry<String, Optional<Integer>>> flattenNestedConfig(
final Map.Entry<String, ConfigValue> entry) {
...
```

tommyulfsparre: Can be made private.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/DurationThresholdConfig.java
```diff
@@ -0,0 +1,89 @@
...
+        .stream()
+        .collect(
+            Collectors.toMap(
+                methodEntry -> String.format("%s:%s", methodEntry.getKey(), entry.getKey()),
+                methodEntry -> Optional.of(methodEntry.getValue())
+            )
+        ).entrySet();
+  }
+
+  public static DurationThresholdConfig parseConfig(final Config config) {
```
tommyulfsparre: I think this function will be easier to read if you remove all temp. variables and move the parsing of `all-endpoints` to its own function. Something like this might work:
```
    if (config.hasPath("endpoint-duration-goal")) {
      final Map<String, Optional<Integer>> goals =
          config.getObject("endpoint-duration-goal")
              .withoutKey("all-endpoints")
              .entrySet()
              .stream()
              .map(DurationThresholdConfig::flattenNestedConfig)
              .flatMap(Collection::stream)
              .collect(
                  Collectors.toMap(
                      Map.Entry::getKey,
                      Map.Entry::getValue));
      return new DurationThresholdConfig(parseAllEndpoints(config),
                                         goals);
    }
    return new DurationThresholdConfig(parseAllEndpoints(config),
                                       ImmutableMap.of());
```

wdyt?

jsferrei: 👍 !

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/DurationThresholdConfig.java
```diff
@@ -0,0 +1,88 @@
...
+        .collect(
+            Collectors.toMap(
+                methodEntry -> String.format("%s:%s", methodEntry.getKey(), entry.getKey()),
+                methodEntry -> Optional.of(methodEntry.getValue())
+            )
+        ).entrySet();
+  }
+
+  private static Optional<Integer> parseAllEndpoints(final Config config) {
+    final Optional<Integer> allEndpoints;
```
tommyulfsparre: nit: No need for these temp. variables:
```
if (config.hasPath("endpoint-duration-goal.all-endpoints")) {
  return Optional.of(config.getInt("endpoint-duration-goal.all-endpoints"));
}
return Optional.empty();
```

tommyulfsparre:
@jsferrei this LGTM but note that the URI parameters is part of the endpoint. 
Example: 
```
GET:/example/<name>
```
which would be nice to document.

dmichel1:
@tommyulfsparre - @jsferrei is out of office for a few days and handed this over to me. I addressed your feedback if you can PTAL again.

Thanks

tommyulfsparre:
@dmichel1 @jsferrei LGTM

dmichel1:
@tommyulfsparre 👍 can we get this released? thanks

tommyulfsparre:
@dmichel1 done!

PR #252, opened 2018-05-15
Extract metric tags from environment variables.

jhaals:
* Bump semantic metrics version
* Add tag extractor that extracts tags from environment variables.
`FFWD_TAG_foo=bar` will result in a kv tag of `foo=bar`.


jhaals:
@tommyulfsparre good point, PTAL if the new README entry is sufficient.

tommyulfsparre:
👍

PR #253, opened 2018-05-15
Update BOM and dependency management

gilles:
The BOM should only contain apollo child modules. As of today it's
impossible to mix the apollo-bom and the jackson-bom in a project
becauseapollo-bom contains jackson-databind -> you end up with both
versions in the classpath and a mess with the maven verifier upper
bound check

All the non apollo dependencies from the BOM are moved to the parent
POM as well as the <distributionManagment> section


gilles:
@tommyulfsparre 

Not sure who maintains apollo but your name pops up in other PR :)

gilles:
Also this is opened on `1.x`. How are you managing `2.x` vs `1.x`, should I open a new PR on `master`?

TheIndifferent:
Made up-to-date PR: https://github.com/spotify/apollo/pull/272

PR #254, opened 2018-05-21
RFC: put error messages in payload of response

bergman:
This should make it easier to see error messages. Using HTTP status reason phrases for communicating anything more than the nature of the error seems to be a bit exotic too.


bergman:
I am sure people have been reading and acting on the reasonPhrase since that has been the only place to get more info on what happened. For backwards compatibility the message could be put in both reasonPhrase and payload but that will just push the problem further into the future.

I expect this change to break a lot of integrations, which is why I put `RFC` in the title.

lepistone:
Closing because of inactivity.

PR #265, opened 2018-11-14
fix Maven warnings about missing maven-surefire-plugin version

mattnworb:
The root pom does not define a version for maven-surefire-plugin, which
this pom seems to assume.

> [WARNING] Some problems were encountered while building the effective model for com.spotify:apollo-bom:pom:1.8.1-SNAPSHOT
> [WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-surefire-plugin is missing. @ com.spotify:apollo-parent:1.8.1-SNAPSHOT, /root/apollo/pom.xml, line 137, column 21

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -137,6 +137,9 @@
             <plugin>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <version>2.22.1</version>
+                <configuration>
+                  <useSystemClassLoader>false</useSystemClassLoader>
+                </configuration>
             </plugin>
             <plugin>
                 <artifactId>maven-failsafe-plugin</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -136,6 +136,7 @@
             </plugin>
             <plugin>
                 <artifactId>maven-surefire-plugin</artifactId>
+                <version>2.22.1</version>
             </plugin>
             <plugin>
                 <artifactId>maven-failsafe-plugin</artifactId>


villespotify:
Hi @mattnworb , we are encountering strange surefire errors as well while working on #266 .

Have you seen https://stackoverflow.com/questions/50661648/spring-boot-fails-to-run-maven-surefire-plugin-classnotfoundexception-org-apache ? In that case, what do you think about adding this:
```
            <configuration>
                <useSystemClassLoader>false</useSystemClassLoader>
            </configuration>
```

mattnworb:
@villespotify sorry I've never seen anything like that and am not familiar :(

villespotify:
Ok, do you mind adding this to the configuration in your PR (this) and see if the build passes?

mattnworb:
@villespotify sure, can't hurt

mattnworb:
@villespotify guess that didn't help. I made this PR since I thought fixing the maven warning noise was simple, but I have no idea why its crashing (and I'm not feeling the pain enough to dig into it yet either).

villespotify:
Ok, thanks @mattnworb - guess it was worth a try at least. Strange that it still fails. We _are_ kind of blocked so may have to investigate further.

tommyulfsparre:
@mattnworb thanks for the PR this is fixed by #266

PR #266, opened 2018-11-19
Bump guava to 26.0

villespotify:
- Update guava version in bom to 26.0-jre
- Fix resulting errors caused by deprecation of addCallback with implicit executor


tommyulfsparre:
LGTM!

PR #269, opened 2018-12-05
[metrics] change default histogram reservoir

dmichel1:
This changes the default histogram reservoir to a [ReservoirWithTtl](https://github.com/spotify/semantic-metrics/tree/9b51f4f6febe7ca251c9410592324645bdb87d6a#histogram-with-ttl).

This eliminates the impact of latencies like p99 being stuck when instances are drained of traffic or the request throughput of a service or endpoint are low.

This is reservoir has already been battle tested internally by being used in Hermes for server metrics.

Would appreciate having this released once merged... we have some people eagerly awaiting this change. 

@tommyulfsparre @jsferrei

### DIFF
--- a/modules/metrics/README.md
+++ b/modules/metrics/README.md
@@ -6,6 +6,10 @@ with the Apollo request/response handling facilities.
 Including this this module in your assembly means that Apollo will add some metrics tracking
 requests per endpoint.
 
+
+A *note* on histograms. This metrics module now uses a [ReservoirWithTtl](https://github.com/spotify/semantic-metrics/tree/9b51f4f6febe7ca251c9410592324645bdb87d6a#histogram-with-ttl) as the **default** reservoir for histograms. This eliminates the impact of latencies like p99 being stuck when instances are drained of traffic or the request throughput of a service or endpoint are low.
+
+
 ## Metrics
 
 All metrics will be tagged with the following tags:
@@ -121,6 +125,7 @@ A Meter, tagged with:
 This meter will only be created if a duration goal is set in the configuration (`endpoint-duration-goal`). The meter will be marked when a request meets its goal.
 
 
+
 ## [ffwd](https://github.com/spotify/ffwd) reporter
 
 The metrics module includes the [ffwd reporter](https://github.com/spotify/semantic-metrics#provided-plugins)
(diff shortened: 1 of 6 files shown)


mpavlov:
👍

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/MetricsConfig.java
```diff
@@ -87,4 +95,8 @@
   public Set<Integer> precreateCodes() {
     return precreateCodes;
   }
+
+  public int reservoirTtl() {
+    return reservoirTtl;
```
tommyulfsparre: We have a helper function in [ConfigUtil](https://github.com/spotify/apollo/blob/1.x/apollo-environment/src/main/java/com/spotify/apollo/environment/ConfigUtil.java#L40) that you can use instead of the conditional above.
 
```suggestion
    return optionalInt("metrics.reservoir-ttl").orElse(DEFAULT_TTL_SECONDS);
```

modules/metrics/README.md
```diff
@@ -132,6 +137,7 @@ key | type | required | note
 --- | ---- | -------- | ----
 `metrics.server` | string list | optional | list of [`What`](src/main/java/com/spotify/apollo/metrics/semantic/What.java) names to enable; defaults to [ENDPOINT_REQUEST_RATE, ENDPOINT_REQUEST_DURATION, DROPPED_REQUEST_RATE, ERROR_RATIO]
 `metrics.precreate-codes` | int list | optional | list of status codes to precreate request-rate meters for, default empty
+`metrics.reservoir-ttl` | int | optional | When to purge old values from the histogram, default 300 seconds.
```
tommyulfsparre: The ttl effects the memory usage on the reservoir, this might be worth mentioning or put an upper bound on.

modules/metrics/src/main/java/com/spotify/apollo/metrics/semantic/MetricsConfig.java
dmichel1: fixed 👍

modules/metrics/README.md
dmichel1: 👍 - I added a note about that.

dmichel1:
@tommyulfsparre I added a test for the behavior we expect. It's the only one I could think of even though its a bit of a duplicate with what is already being tested in `SemanticMetrics`. Let me know if you had thoughts on what else to could be tested.

PR #270, opened 2018-12-07
Automatic-Module-Name added to all manifests.

TheIndifferent:
As described by http://branchandbound.net/blog/java/2017/12/automatic-module-name/ , adding `Automatic-Module-Name` allows us to move to Java 11 without worrying that the jar file name will change.

### DIFF
--- a/apollo-api-impl/pom.xml
+++ b/apollo-api-impl/pom.xml
@@ -120,6 +120,19 @@
                 <filtering>true</filtering>
             </resource>
         </resources>
+        <plugins>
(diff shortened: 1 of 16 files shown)


TheIndifferent:
@klaraward no reason, and no strong feelings about it. Was considering either using package name or just making module name as descriptive as possible, went for latter, but I can change if anyone prefers root package name approach more.

TheIndifferent:
Seems like current version of Apollo is hugely incompatible with JPMS any way...

PR #274, opened 2019-02-14
Enable auto discovery

jwiklund:
All modules marked @AutoService(ApolloModule.class) must have a public constructor or boot will fail when running with enabled discovery.


modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsModule.java
```diff
@@ -65,8 +65,8 @@
 
   private static final Logger LOG = LoggerFactory.getLogger(MetricsModule.class);
 
-  private MetricsModule() {
-    // prevent external instantiation
+  // Visible for SPI support
```
klaraward:
```suggestion
  // Should not be used, only here to be visible for SPI support.
```

PR #276, opened 2019-02-19
Update comment for public constructor

klaraward:
Update comment to indicate that public constructor should not be used except by SPI
Similar to change in #274


tommyulfsparre:
👍

PR #280, opened 2019-06-20
Bump version of okio

dmichel1:
A later version of okio is used within a tracing library to avoid the need for pinning the dependency bump the version to match what that library uses.


lepistone:
Thanks! Superseded by #285

dmichel1:
@lepistone 👍

PR #284, opened 2019-08-23
Update http client module example

spkrka:
Since HttpClient is not a public class, rewrite the example to use the public Client interface.

### DIFF
--- a/modules/okhttp-client/README.md
+++ b/modules/okhttp-client/README.md
@@ -25,12 +25,14 @@ key | type | required | note
 
 ```java
 public static void main(String[] args) throws Exception {
+
   Service service = Services.usingName("test")
       .withModule(HttpClientModule.create())
+      .withModule(ApolloEnvironmentModule.create())
       .build();
 
   try (Service.Instance instance = service.start(args)) {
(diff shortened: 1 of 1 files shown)


modules/okhttp-client/README.md
```diff
@@ -25,12 +25,14 @@ key | type | required | note
...
 public static void main(String[] args) throws Exception {
+
   Service service = Services.usingName("test")
       .withModule(HttpClientModule.create())
+      .withModule(ApolloEnvironmentModule.create())
       .build();
 
   try (Service.Instance instance = service.start(args)) {
-    HttpClient httpClient = instance.resolve(HttpClient.class);
+    Client httpClient = instance.resolve(ApolloEnvironment.class).environment().client();
```
danielnorberg: Would it work to just `instance.resolve(Client.class)` instead?

spkrka: I don't think so - HttpClientModule does this: https://github.com/spotify/apollo/blob/1.x/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpClientModule.java#L43

It neither binds Client nor IncomingRequestAwareClient so you can't resolve those.

PR #293, opened 2019-10-30
Fix exception thrown when restarting ServiceHelper

mklamra:
The Maven Failsafe plugin supports rerunning failed tests (http://maven.apache.org/surefire/maven-failsafe-plugin/integration-test-mojo.html#rerunFailingTestsCount). A rerun causes the Failsafe plugin to restart all test rules. If you enable reruns for a test using a `ServiceHelper` as a test rule, the `ServiceHelper` will throw an `IllegalStateException` during the restart due to https://github.com/spotify/apollo/blob/1.x/apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelper.java#L534. This happens because the `started` `CountDownLatch` is not reset during the shutdown. This PR fixes the issue by changing `started` to a `Semaphore` that is reset during shutdown.

### DIFF
--- a/apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelper.java
+++ b/apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelper.java
@@ -62,10 +62,10 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 import java.util.concurrent.CompletionStage;
-import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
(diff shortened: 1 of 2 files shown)


toweroy:
@mklamra change LGTM, could we add a simple test for a fresh start of the `ServiceHelper`? just to test that the `instanceCreated` does what it's meant to do if it is to run for the first time.

mklamra:
@toweroy Isn't that already covered by the `smokeTest` and `smokeTest2` tests?

toweroy:
@mklamra totally right, missed those.

mklamra:
Thanks @toweroy! Could you merge and release the change? I don't have push access to the repo.

PR #298, opened 2019-12-19
Add JUnit 5 ServiceHelper extension

lutovich:
This PR is a combination of 3 commits:

#### Use JUnit 5 with vintage and jupiter engines

Migrate from JUnit 4 to JUnit 5 with a backward-compatible test engine. This allows writing both JUnit 4 and 5 tests. It also makes it possible to add a JUnit 5 ServiceHelper extension.

#### Bump cucumber version

Move from deprecated 1.2.5 to 4.8.0 primarily to get rid of deprecation warnings and be more JUnit 5 compatible.

#### Add JUnit 5 service helper extension

The added extension mirrors `ServiceHelper` rule API and allows to access the created helper via a getter or an injected method parameter.

Declarative registration of this extension is not allowed and throws an exception. This is because there is no good way to configure the extension and specify `AppInit` or `ServiceHelper` configuration.

### DIFF
--- a/apollo-test/src/main/java/com/spotify/apollo/test/ServerHelperSetup.java
+++ b/apollo-test/src/main/java/com/spotify/apollo/test/ServerHelperSetup.java
@@ -1,0 +1,45 @@
+/*
+ * -\-\-
+ * Spotify Apollo Testing Helpers
+ * --
+ * Copyright (C) 2013 - 2015 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+package com.spotify.apollo.test;
+
+import com.spotify.apollo.module.ApolloModule;
+
+interface ServerHelperSetup<T extends ServerHelperSetup<T>> {
+
+  T domain(String domain);
+
+  T disableMetaApi();
+
+  T args(String... args);
+
+  T conf(String key, String value);
+
+  T conf(String key, Object value);
+
+  T resetConf(String key);
+
+  T forwardingNonStubbedRequests(boolean forward);
+
+  T startTimeoutSeconds(int timeoutSeconds);
+
+  T withModule(ApolloModule module);
+
+  T scheme(String scheme);
+}
--- a/apollo-api-impl/pom.xml
+++ b/apollo-api-impl/pom.xml
@@ -86,8 +86,8 @@
 
         <!--test scope-->
         <dependency>
-            <groupId>junit</groupId>
-            <artifactId>junit</artifactId>
+            <groupId>org.junit.vintage</groupId>
+            <artifactId>junit-vintage-engine</artifactId>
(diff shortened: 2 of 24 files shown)


apollo-test/src/main/java/com/spotify/apollo/test/ServerHelperSetup.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * -\-\-
+ * Spotify Apollo Testing Helpers
+ * --
+ * Copyright (C) 2013 - 2015 Spotify AB
```
klaraward: Nitpick, guess the copyright end year should be updated?

klaraward:
The only change someone using this new version would need to do is change from cucumber.api to io.cucumber in their imports? Or specify the version of the old cucumber themselves?

apollo-test/src/main/java/com/spotify/apollo/test/ServerHelperSetup.java
lutovich: Oh, right I haven't noticed this

lutovich:
@klaraward as far as I understand, the change of cucumber version is purely internal. It is only used for acceptance tests in `apollo-http-service` module and dependency has `test` scope. This PR should be purely an additive change from a user perspective.

lutovich:
Hi @lepistone,

Sorry for the delayed response!

#### Dependency versions:
 1. `org.junit.jupiter:junit-jupiter-engine:5.5.2` - needed to execute JUnit 5 tests in Apollo and add `ServiceHelperExtension`
 1. `org.junit.vintage:junit-vintage-engine:5.5.2` - needed to execute JUnit 4 tests in Apollo and keep `ServiceHelper` test rule
 1. `org.junit.platform:junit-platform-testkit:1.5.2` - needed to test JUnit 5 `ServiceHelperExtension`
 1. `io.cucumber:cucumber-java:4.8.0` - upgraded from `info.cukes:cucumber-java:1.2.5` needed to execute acceptance tests in `apollo-http-service` module
 1. `io.cucumber:cucumber-junit:4.8.0` - upgraded from `info.cukes:cucumber-junit:1.2.5` needed to execute acceptance tests in `apollo-http-service` module

All of them were added to the `dependencyManagement` section of the parent pom with the default `compile` scope.

The scope is always overridden to be `test` for `junit-platform-testkit`, `cucumber-java` and `cucumber-junit`. So these are purely test-only dependencies for Apollo itself. Test dependencies are not transient. If someone depends on Apollo `test-jar`s (not a good thing to do) then they will not automatically get any of those. That's why I think these 3 dependencies should not bring any trouble.

`junit-jupiter-engine` and `junit-vintage-engine` remain `compile` scope dependencies for module `apollo-test`. Where `junit-vintage-engine` replaces JUnit 4 dependency. This could be problematic because a test dependency on `apollo-test` would transitively pull in two test engines of some fixed version. Before this PR, this was also the case but for `junit:junit:4.12` which is updated very infrequently so the problem was smaller.
To solve this problem we could set the scope for jupiter and vintage engine dependencies to `provided`. This makes these dependencies not transitive. What do you think about this solution?

#### Dependency versions in root POMs:
Looks like neither `com.spotify:foss-root` not `com.spotify:root` define `demendencyManagement`. I guess that's to make projects responsible for their own dependencies. So we probably do not need to modify those.

#### Documentation:
It would be nice to modify our documentation and Golden Paths to use JUnit 5. Some examples there already use it but integration tests don't. They use JUnit 4 `ServiceHelper` rule.

Also, I'll add javadoc to `ServiceHelperExtension` if this PR will be merged.

#### Deprecations:
IMO we do not need to deprecate anything, maybe only add a notice to `ServiceHelper` javadoc saying that a JUnit 5 extension is available. JUnit 4 is probably not going anywhere for a long time. Deprecating `ServiceHelper` will introduce lots of warnings for Apollo users.

Hope these thoughts make some sense 🙂

PR #302, opened 2020-02-05
Use favicon from the new CDN

rnhsp:


klaraward:
So if I understand this correctly, this fix is for the website http://spotify.github.io/apollo/
and someone would deploy it manually using instructions in https://github.com/spotify/apollo/tree/1.x/website

rnhsp:
Yes, this is for the https://spotify.github.io/apollo/ website.

Ah, I see, I wasn't aware of the build process described in https://github.com/spotify/apollo/blob/master/website/README.md. I'll close this and make the change in the correct place (https://github.com/spotify/apollo/blob/master/website/source/_layouts/default.ejs), which would mean that build process should work as expected. :)

rnhsp:
https://github.com/spotify/apollo/pull/303

PR #305, opened 2020-03-10
Add ERROR_RATIO_4XX to default set of metrics

bergetp:
Emit ERROR_RATIO_4XX metric by default and update documentation.


klaraward:
Just a note that the corresponding 5xx metrics were added in this PR - https://github.com/spotify/apollo/pull/301

PR #306, opened 2020-03-15
Add Automatic-Module-Name manifest entries to jars

rouzwawi:
To enable publishing libraries using JPMS

### DIFF
--- a/apollo-api-impl/pom.xml
+++ b/apollo-api-impl/pom.xml
@@ -131,6 +131,20 @@
                 <filtering>true</filtering>
             </resource>
         </resources>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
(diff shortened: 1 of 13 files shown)


apollo-api-impl/pom.xml
```diff
@@ -131,6 +131,20 @@
...
             </resource>
         </resources>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifestEntries>
+                            <!-- note: apollo-api-impl splits some packages with apollo-api -->
```
klaraward: So this would really mean thing would have to be moved around? IIRC packages are not supposed to be in multiple modules?

rouzwawi: Yes eventually. I think for now there's ways around it, either by not depending on two modules that split a package, or keep those dependencies on the classpath.

PR #307, opened 2020-04-15
Fix ServiceHelperExtension::conf

mishako:
It was using the key argument as both key and value.

### DIFF
--- a/apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelperExtension.java
+++ b/apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelperExtension.java
@@ -183,7 +183,7 @@ public final class ServiceHelperExtension
 
   @Override
   public ServiceHelperExtension conf(String key, Object value) {
-    serviceHelper.conf(key, key);
+    serviceHelper.conf(key, value);
     return this;
   }
 
--- a/apollo-test/src/test/java/com/spotify/apollo/test/ServiceHelperExtensionTest.java
+++ b/apollo-test/src/test/java/com/spotify/apollo/test/ServiceHelperExtensionTest.java
@@ -19,35 +19,27 @@
  */
 package com.spotify.apollo.test;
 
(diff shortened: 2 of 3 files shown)


mishako:
The diff looks weird because I renamed `ServiceHelperExtensionTest` to `ServiceHelperExtensionUsageTest`.

klaraward:
Curious how you discovered this and what bugs it caused?

mishako:
@klaraward I was migrating a test to junit5, switched from `ServiceHelper` to `ServiceHelperExtension` and it started failing. It was easy to fix on my side though: by converting my int value to a string.

apollo-test/src/main/java/com/spotify/apollo/test/ServiceHelperExtension.java
```diff
@@ -183,7 +183,7 @@ public ServiceHelperExtension conf(String key, String value) {
 
   @Override
   public ServiceHelperExtension conf(String key, Object value) {
-    serviceHelper.conf(key, key);
+    serviceHelper.conf(key, value);
```
lutovich, line 186: 😳

apollo-test/src/test/java/com/spotify/apollo/test/ServiceHelperExtensionTest.java
```diff
@@ -19,35 +19,27 @@
  */
 package com.spotify.apollo.test;
 
-import static org.junit.jupiter.engine.descriptor.JupiterEngineDescriptor.ENGINE_ID;
-import static org.junit.platform.engine.discovery.DiscoverySelectors.selectClass;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertThat;
```
lutovich, line 23: Maybe use `org.hamcrest.MatcherAssert#assertThat()` instead?
`org.junit.Assert#assertThat()` is from JUnit 4 and this is a JUnit 5 test

lepistone: Good point! I'll merge anyway. PR very welcome!

PR #310, opened 2020-06-15
Remove service-framework-version tag from metrics

bergetp:
This tag changes with every new release of Apollo. Due to limitations with an internal systems we would like the metric tags to stay the same when upgrading to a new version of Apollo.


bergetp:
@malish8632 seems I don't have access to merge this..

malish8632:
I don't have permissions as well. I pinged internally our team.
Thanks for flagging.

PR #312, opened 2020-07-09
Update hamcrest to 2.2 and remove deprecated modules

gilles:


gilles:
@lepistone

gilles:
thanks for taking a look!

I made it my sort of quixotesque mission to remove references to `-library` `-core` everywhere I can, It's silly I know. I've been able to compile with jdk8  at some point. I'll take another look. Are you open to `<exclusions>` or rather add the deprecated ones where needed?

lepistone:
@gilles going after unnecessary dependencies is great, not silly! But in this particular case, I think the problem is that junit 4 depends on hamcrest-core (or -library, or both). So the end result is that by removing them, we get a mix of 2.2 and 1.x. So these empty 2.2 -core and -library packages are one possible fix.
But if you prefer exclusions, that should also work and I'd happy to do that as well.

gilles:
Looking at the failure it does not look like junit or hamcrest related, more like a transient thing (I can build locally).

Regardless I think this is a cleaner way: remove all `-core`, `-library` references in `<dependency>` but keep in the root `<dependencyManagement>` so it resolves transitive deps while not exposing the deprecated things when exporting apollo

PR #313, opened 2020-08-05
Remove jetty service and examples

lepistone:
These are not much used in Spotify, who are likely to be the main users of the Apollo framework. The examples are very old and don't
necessarily represent the usage of Apollo today. The Jetty service and
the related HTTP assembly were good examples of usage of Apollo Open
Source, but nowadays, not having been updated in a while, they risk
creating more confusion than value.


tommyulfsparre:
@lepistone thanks!

PR #316, opened 2020-08-13
upgrade okio to 2.7.0

mattnworb:
A number of internal Spotify libraries related to distributed tracing are introducing a requirement on okio 2.x and okhttp3 4.x.

Between 1.x and 2.x, okio was rewritten from Java to Kotlin, and promises to be API compatible with 1.x:

> Okio 2 is a major release that upgrades the library’s implementation language from Java to Kotlin.

> Okio 2.x is binary-compatible with Okio 1.x and does not change any behavior. Classes and .jar files compiled against 1.x can be used with 2.x without recompiling.

> Okio 2.x is .java source compatible with Okio 1.x in all but one corner case. In Okio 1.x Buffer would throw an unchecked IllegalStateException when attempting to read more bytes than available. Okio 2.x now throws a checked EOFException in this case. This is now consistent with the behavior of its BufferedSource interface. Java callers that don’t already catch IOException will now need to.

https://square.github.io/okio/changelog/#version-200-rc1

Since Apollo uses okio, this upgrades Apollo to the 2.x line of okio as a way to verify that there are no problems within Apollo before we start to upgrade Apollo-using services.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
             <dependency>
                 <groupId>com.squareup.okio</groupId>
                 <artifactId>okio</artifactId>
-                <version>1.17.4</version>
+                <version>2.7.0</version>
             </dependency>
             <dependency>
                 <groupId>com.google.inject</groupId>


klaraward:
What about okhttp, does that need also to be changed?

mattnworb:
@klaraward no. The apollo-okhttp module is using okhttp v2. In v3 they changed the Java package name and Maven artifactId to “okhttp3” for API compatibility reasons, which allows both okhttp and okhttp3 to both exist on the classpath without a conflict. This means Apollo isn’t using okhttp3 today so it should not be affected by upgrading okhttp3 from v3 to v4. 

https://square.github.io/okhttp/changelog_3x/#version-300-rc1
https://jakewharton.com/java-interoperability-policy-for-major-version-updates/

mattnworb:
(It would be good to eventually upgrade to okhttp3 but not as a part of this PR)

erikjoh (approved):

PR #317, opened 2020-08-20
Update README for apollo-entity

benkogan:
This PR fixes the bulleted list indentation for the apollo-entity README. (Previously, the indented items didn't render at the correct indentation level — this confused me the first time I read through!) It also tweaks a couple words/punctuation for consistency.

### DIFF
--- a/apollo-entity/README.md
+++ b/apollo-entity/README.md
@@ -1,11 +1,11 @@
 Apollo Entity
 =============
 
-A set of Apollo Middleware for working with entity types in route handlers.
+A set of Apollo middleware for working with entity types in route handlers.
 
 ## Why
 
-A common pattern when writing route handlers is to serialize and deserialize to some API level
+A common pattern when writing route handlers is to serialize and deserialize to some API-level
 entity types. Writing such (de)serialization code quickly becomes repetitive and always looks
 the same.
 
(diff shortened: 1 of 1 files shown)


apollo-entity/README.md
```diff
@@ -29,7 +29,7 @@ However, this does not work. Theres nothing handling how the `Person` type is be
 from a `ByteString` which is the payload type Apollo works with.
 
 At this point, you'll start writing some custom middlewares using something like Jackson. This
-library contains exactly those middlewares, ready to use directly with your routes.
+library contains exactly those middleware, ready to use directly with your routes.
```
klaraward, line 32: this change made the grammar incorrect....

benkogan: Hi Klara! I was curious about this and could not find a definitive reference, but Googling "is middleware plural?" I see conflicting answers — it's either already plural (like _software_) or it's pluralized as _middlewares_ (which, for what it's worth, the macOS spell checker does not like).

I went with the former since it's consistent with the opening line of the document, where it's used in plural as _middleware_:
> A set of Apollo Middleware for working with entity types in route handlers.

but I'm happy to revert all changes beyond the bullet indentation fix, if you'd prefer!

klaraward: I'm happy as long as it is consistent with the rest of the text. Maybe middleware is the plural form..

klaraward:
Replaced by #358

PR #320, opened 2020-09-01
Allow loading extra modules for a Service

mspringare:
Add possibility to specify extra set of modules that will always be loaded


apollo-core/src/test/java/com/spotify/apollo/core/ServiceImplTest.java
```diff
@@ -578,6 +579,39 @@ public void testSimulateCtrlCWaitingForShutdown() throws Exception {
...
   }
 
+  @Test
+  public void testExtraModules() throws Exception {
+    Service service = ServiceImpl.builder("test")
+        .build();
+
+    try (Service.Instance instance = service.start(
+        new String[0], ConfigFactory.empty(), ImmutableSet.of(new NoopModule()))) {
+      NoopModule.Noop noop =
```
tommyulfsparre: @patriknordlen @mspringare can we reuse the `CountingModuleWithPriority` here and remove the `NoopModule` all togheter?

patriknordlen: @tommyulfsparre technically we could of course but personally I think it's cleaner to keep them separate as we're using them for different purposes.

apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
```diff
@@ -232,14 +236,20 @@ InstanceImpl initInstance(
...
+            modules,
+            extraModules.stream()
+                .filter(m -> modules.stream().noneMatch(am -> m.getId().equals(am.getId())))
+                .collect(Collectors.toSet()));
 
     if (moduleDiscovery) {
-      allModules = Sets.union(modules, ImmutableSet.copyOf(ServiceLoader.load(ApolloModule.class)));
-    } else {
-      allModules = modules;
+      return Sets.union(allModules, ImmutableSet.copyOf(ServiceLoader.load(ApolloModule.class)));
```
klaraward: We're not doing the deduping against the modules loaded visa ServiceLoader, but I guess that's ok :)

toweroy (approved):

klaraward (approved):

apollo-core/src/main/java/com/spotify/apollo/core/Service.java
```diff
@@ -78,24 +77,43 @@
...
+   *     ApolloModule#getId()}.
+   * @return a new instance of this service that is up and running.
+   * @throws ApolloHelpException if the user wants to show command-line help and not start the
+   *     application.
+   * @throws ApolloCliException if something else related to CLI parsing failed.
+   * @throws java.io.IOException if the application could not start for some other reason.
    */
+  default Instance start(String[] args, Config config, Set<ApolloModule> extraModules)
+      throws IOException {
+    return this.start(args, config);
```
erikjoh: Not sure I understand why this default implementation is required.
This default behavior does not match what the docstring says.
I.e. if nothing overrides this the `extraModules` are ignored

erikjoh: Discussion took place in separate communication channel.
I'm fine with whatever :slightly_smiling_face:

erikjoh (approved):

erikjoh (approved):

erikjoh (approved):

PR #326, opened 2021-01-21
[metrics] Add flush option to ffwd reporters

lmuhlha:
We received a report that: 

"In a high throughput pubsub subscriber/publisher, every deploy looks like it is losing a lot of messages during shutdown on exit.
If you prematurely shut down the publisher before the subscriber, a lot of messages get dropped (with failure reported), but they never showed up in the metrics.
Should calling FastForward(Http)Reporter.stop() be enough to flush the data? Do you think the problem would be that Apollo doesn’t do that, or that the stop method itself doesn’t flush?
Neither apollo nor ffwd was doing a final flush of these metrics. 
The ffwd reporters are responsible for flushing the metrics. It's running on a 30 second interval between flushes
during shutdown, the ffwd reporter is simply stopped (thread responsible for flushing is shutdown immediately)."
 
This PR allows you to configure both the FastForwardHttpReporter and FastForwardReporter to do a final flush of metrics on shutdown using the `ffwd.flush` flag in your config. This should help remediate the loss of metrics in cases like what I described above. 

I ran into a few issues when upgrading the libraries for this PR, leading me to make some additions to the parent pom to get this to pass for java 8. The `missinglink` profile was complaining about pretty much every dependency related to the Netflix load balancer used in ffwd-http-reporter. This was remediated by changing the scopes: `<includeScopes>compile,test,runtime</includeScopes>`. Is the `missinglink` profile still something the owners care about? It seems a bit outdated and looks like it only gets called for java8.

### DIFF
--- a/modules/metrics/README.md
+++ b/modules/metrics/README.md
@@ -157,6 +157,7 @@ will then emit a '0' value until the first time a certain status code shows up.
 key | type | required | note
 --- | ---- | -------- | ----
 `ffwd.interval` | int | optional | interval in seconds of reporting metrics to ffwd; default 30
+`ffwd.flush` | boolean | optional | include a final flush of metrics on service shut down; default `False`
 `ffwd.host` | string | optional | host where the ffwd agent is running. default `localhost`
 `ffwd.port` | int | optional | port where the ffwd agent is running. default `19091`
 
@@ -174,6 +175,7 @@ ffwd.port = 19091
 key | type | required | note
 --- | ---- | -------- | ----
 `ffwd.interval` | int | optional | interval in seconds of reporting metrics to ffwd; default 30
+`ffwd.flush` | boolean | optional | include a final flush of metrics on service shut down; default `False`
 `ffwd.discovery.type` | string | required | indicates how to discovery http endpoints. Available options are `static` and `srv`. See below for details.
 
 #### Example
--- a/modules/metrics/src/main/java/com/spotify/apollo/metrics/FfwdConfig.java
+++ b/modules/metrics/src/main/java/com/spotify/apollo/metrics/FfwdConfig.java
@@ -19,6 +19,7 @@
  */
 package com.spotify.apollo.metrics;
 
+import static com.spotify.apollo.environment.ConfigUtil.optionalBoolean;
 import static com.spotify.apollo.environment.ConfigUtil.optionalInt;
 import static com.spotify.apollo.environment.ConfigUtil.optionalString;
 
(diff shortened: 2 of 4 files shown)


modules/metrics/src/main/java/com/spotify/apollo/metrics/FfwdConfig.java
```diff
@@ -102,6 +111,14 @@ int getInterval() {
       host.ifPresent(builder::host);
       port.ifPresent(builder::port);
 
+      if (flush) {
+        return () -> {
+          final FastForwardReporter reporter = builder.build();
+          reporter.start();
+          return reporter::stopWithFlush;
```
irock: How about moving the flag to this line, and merge this lambda with the one below for less code duplication?

```
return () -> {
        final FastForwardReporter reporter = builder.build();
        reporter.start();
        return flush ? reporter::stopWithFlush : reporter::stop;
      };
```

irock (approved):

pom.xml
```diff
@@ -602,6 +603,10 @@
                                 <ignoreDestinationPackage>
                                     <package>com.jcraft.jzlib</package>
                                 </ignoreDestinationPackage>
+                                <!-- ignore references to Android util in okhttp3 -->
+                                <ignoreDestinationPackage>
+                                    <package>android.util</package>
```
klaraward, line 608: If you think this is a "safe" thing to ignore, which I assume it is, I'm totally fine with this.

klaraward (approved):

modules/metrics/README.md
```diff
@@ -157,6 +157,7 @@ will then emit a '0' value until the first time a certain status code shows up.
 key | type | required | note
 --- | ---- | -------- | ----
 `ffwd.interval` | int | optional | interval in seconds of reporting metrics to ffwd; default 30
+`ffwd.flush` | boolean | optional | include a final flush of metrics on service shut down; default `False`
```
tommyulfsparre, line 160: Is there a reason to make this configurable as flushing metrics on shutdown seems like the behaviour we always want?

tommyulfsparre:
@lmuhlha could we add some test(s) for this?

lmuhlha:
@tommyulfsparre added some tests here: https://github.com/spotify/apollo/pull/326/files#diff-d53750fc42fc89f48dad915eec502ac8a744008da82f276a28def2a170b4a886

Was there anything else you had in mind? (There are tests for the functionality in the semantic metrics repo as well: https://github.com/spotify/semantic-metrics/pull/96)

modules/metrics/README.md
lmuhlha: The idea was not to force new behavior on users that don't have a problem with the way it was working previously. 
We could make it the default though.

lmuhlha: Changed this to default to "True". Users can still opt out to the previous functionality if they'd like with the flag set to "False". It adds a block to shutdown so some people might care.

tommyulfsparre:
@lmuhlha there is no test that covers that `reporter::stopWithFlush` gets called on shutdown. Seeing that there is no existing test that actually tests that the reporter is even started and the code is not easily testable we will not block on that.

modules/metrics/README.md
tommyulfsparre: Good point regarding blocking shutdown. 
I guess that is only relevant for the HttpReporter as the other uses UDP? 
How long will the HttpReporter block if the metrics api would be unavailable ?

lmuhlha:
Thank you!

modules/metrics/README.md
lmuhlha: Internally, there are very few  (if any) people using the http reporter. We've been debating deprecating it and removing it. It doesn't block if metricsapi is unavailable. It just takes a bit longer for shutdown since it's trying to run that final flush of metrics, as opposed to previously it would immediately send shutdown and nothing else. 

The remaining issue is that ffwd would need to wait a bit to really get *all* the metrics, but since it's part of the same pod it will get a sigterm when the main application does. A more guaranteed flush of all the metrics would probably require some kind of wait or timeout.

PR #328, opened 2021-02-02
remove two empty Java source files with funny extensions

mattnworb:
These both seem to have been created in 25e2417 - probably an accident from an automated refactor or rename.


erikjoh:
Thanks!

PR #330, opened 2021-02-11
Upgrade semantic metrics to 1.1.7

lmuhlha:
Latest release: https://github.com/spotify/semantic-metrics/releases/tag/1.1.7

**Note**: This is a breaking change if you were querying for sprawling keys (ex: apollo.apollo.apollo.apollo). You may have to update your queries to use the non-sprawling key. If you are not using a key in your query, the results should stay the same.

Would be good to add that note to the release notes for apollo as well :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -46,7 +46,7 @@
         <auto-value.version>1.7.4</auto-value.version>
         <logback.version>1.2.3</logback.version>
         <jackson.version>2.11.2</jackson.version>
-        <semantic-metrics.version>1.1.6</semantic-metrics.version>
+        <semantic-metrics.version>1.1.7</semantic-metrics.version>
         <guava.version>29.0-jre</guava.version>
         <netty.version>4.1.45.Final</netty.version>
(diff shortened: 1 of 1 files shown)


klaraward:
@lmuhlha  Is the usage of sprawling keys common, how would you find it in codebase?

lmuhlha:
Screen Shot 2021-02-16 at 10 40 23 AM
Checking with bigquery on the heroic query dataset, there are about 1039 queries with `apollo.apollo` as they key in a day. Anything more than that (like `apollo.apollo.apollo`, `apollo.apollo.apollo.apollo`, etc) doesn't seem to be queried on average. Since there is a `prefix` in a key, `apollo.apollo` may just be an accurate key where apollo has the prefix set to `apollo` followed by a normal `apollo` key. So some, if not all, might not be changed in the case of the 1039 queries. It's just something people should be aware of when upgrading as a precaution.

lmuhlha:
In reference to Grafana, it's this: 
Screen Shot 2021-02-16 at 10 45 17 AM

PR #331, opened 2021-02-25
change auto-matter dependency scope to provided

mattnworb:
I noticed in a project that depends on `apollo-api-impl` that
auto-matter and its transitive dependencies (like javapoet) were being
pulled into the project as compile-scope dependencies.


klaraward:
Will have a look

PR #332, opened 2021-02-26
Remove io.netty dependency management

klaraward:
Added in https://github.com/spotify/apollo/pull/279, the only thing that uses netty is the mockserver which is used in the test scope.


lepistone (approved):
Great you cleaned it up, thank you!

PR #335, opened 2021-03-26
add http metric module

ditualina:
Add http client decorator to log outgoing requests

### DIFF
--- a/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
+++ b/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
@@ -56,7 +56,7 @@ public class HttpMetricModule extends AbstractApolloModule {
     }
 
     public IncomingRequestAwareClient apply(IncomingRequestAwareClient baseClient) {
-      return new LoggingHttpClient(baseClient, this.metricsRegistry);
+      return new MetricsHttpClient(baseClient, this.metricsRegistry);
     }
   }
 }
--- a/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
+++ b/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
@@ -2,7 +2,7 @@
  * -\-\-
  * Spotify Apollo okhttp Client Module
  * --
- * Copyright (C) 2013 - 2021 Spotify AB
+ * Copyright (C) 2021 Spotify AB
  * --
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
+++ b/modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
@@ -1,0 +1,62 @@
+/*-
+ * -\-\-
+ * Spotify Apollo okhttp Client Module
+ * --
+ * Copyright (C) 2013 - 2021 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+package com.spotify.apollo.http.client;
+
+import com.google.inject.Inject;
+import com.google.inject.multibindings.Multibinder;
+import com.spotify.apollo.environment.ClientDecorator;
+import com.spotify.apollo.environment.IncomingRequestAwareClient;
+import com.spotify.apollo.module.AbstractApolloModule;
+import com.spotify.apollo.module.ApolloModule;
+import com.spotify.metrics.core.SemanticMetricRegistry;
+
+public class HttpMetricModule extends AbstractApolloModule {
+
+  public static ApolloModule create() {
+    return new HttpMetricModule();
+  }
+
+  @Override
+  public String getId() {
+    return "http-metric-module";
+  }
+
+  @Override
+  protected void configure() {
+    Multibinder.newSetBinder(this.binder(), ClientDecorator.class)
+        .addBinding()
+        .to(HttpMetricClientDecorator.class);
+  }
+
+  public static class HttpMetricClientDecorator implements ClientDecorator {
+
+    private final SemanticMetricRegistry metricsRegistry;
+
+    @Inject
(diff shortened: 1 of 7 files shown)


modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/LoggingHttpClient.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private void logOutgoingRequest(
+      final Request request, final long elapsedMillis, @Nullable final Response<?> response) {
+    final String statusCode =
+        response != null ? Integer.toString(response.status().code()) : "EXCEPTION";
+
+    final String targetHost = getHost(request);
+
+    metricRegistry
```
gilles: should probably use a ` MetricIdCache`

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/LoggingHttpClient.java
```diff
@@ -0,0 +1,107 @@
...
+                .tagged("what", "request-latency")
+                .tagged("unit", "ms")
+                .tagged("protocol", "http/https")
+                .tagged("target-host", targetHost))
+        .update(elapsedMillis);
+  }
+
+  private String getHost(Request request) {
+    try {
+      return URI.create(request.uri()).getHost();
```
mattnworb: I think this could just be `request.httpUrl().host()`, without having to reparse a URL or handle an exception

https://github.com/square/okhttp/blob/98a2ce65a424a700cf08d76df9d8e7a4afe42eff/okhttp/src/main/java/com/squareup/okhttp/HttpUrl.java#L390

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/LoggingHttpClient.java
```diff
@@ -0,0 +1,107 @@
...
+    final long start = System.currentTimeMillis();
+
+    return client
+        .send(request, incoming)
+        .whenComplete(
+            (response, throwable) -> {
+              final long elapsedMillis = System.currentTimeMillis() - start;
+              if (throwable != null) {
+                logOutgoingRequest(request, elapsedMillis, null);
+                logger.error("Exception caught: ", throwable);
```
mattnworb: typically when you have a client class like this which wraps another instance, you shouldn't log exceptions that will be logged by the code using this also - I think logging this here means its likely a single exception will be logged multiple times.

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/LoggingHttpClient.java
ditualina: Hmm, but `Request` interface has no direct `httpUrl()` so it will still need to be built

mattnworb: Sorry, I thought this was the okhttp `Request` class 🤦‍♂️  please disregard

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/LoggingHttpClient.java
```diff
@@ -0,0 +1,107 @@
...
+import com.spotify.metrics.core.SemanticMetricRegistry;
+import java.net.URI;
+import java.util.Optional;
+import java.util.concurrent.CompletionStage;
+import javax.annotation.Nullable;
+import okio.ByteString;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LoggingHttpClient implements IncomingRequestAwareClient {
```
klaraward: The name of the class, even though it's not exposed to the user, indicates that this is adding logging, when it's actually adding metrics?

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/LoggingHttpClient.java
ditualina: Hmm, not sure if it's needed, saw the recommendation is to use this if there are many instances created, but I don't think so far we have that many uses cases 🤔 
https://github.com/spotify/semantic-metrics#metricidcache

ditualina:
@klaraward build fails due to a mockito error, possible explanation is the `java-mockito` compatibility : https://stackoverflow.com/a/59888581/4475136

Since we have this running for `jdk8`, `jdk11` and `jdk13`, I wonder how we could solve this

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
```diff
@@ -0,0 +1,62 @@
+/*-
+ * -\-\-
+ * Spotify Apollo okhttp Client Module
+ * --
+ * Copyright (C) 2013 - 2021 Spotify AB
```
klaraward: Copyright should be just 2021, right?

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
```diff
@@ -0,0 +1,62 @@
...
+
+import com.google.inject.Inject;
+import com.google.inject.multibindings.Multibinder;
+import com.spotify.apollo.environment.ClientDecorator;
+import com.spotify.apollo.environment.IncomingRequestAwareClient;
+import com.spotify.apollo.module.AbstractApolloModule;
+import com.spotify.apollo.module.ApolloModule;
+import com.spotify.metrics.core.SemanticMetricRegistry;
+
+public class HttpMetricModule extends AbstractApolloModule {
```
klaraward, line 31: Can you also add something to the readme for this module, to explain how to use this? Do you use both this module and the regular module, or just the new one?

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/LoggingHttpClient.java
klaraward: How could you tell if a cache was needed, just base on the clients performance?

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
ditualina: Hmm, I guess. this was autogenerated when I ran `mvn license:update-file-header`

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
ditualina: both, this just adds the metrics and should not replace the regular one. I'll update the readme 👍

modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/LoggingHttpClient.java
ditualina: yes, so far, this implementation was used in 3 of our services and we didn't see any performance affected. Also this will be a separate module and not included in the HttpClientModule. But if you think it makes sense using `MetricIdCache`, I could change it

klaraward: I think you can add it later if needed.

klaraward (approved):

PR #323, opened 2020-10-26
Bump junit-bom to 5.7.0

honnix:
See issue here: https://github.com/junit-team/junit4/issues/1678

PR #336, opened 2021-04-13
Update junit dependencies.

rulle-io:
As a result of this PR junit:junit:4.12 (managed) =>  junit:junit:4.13.1 (managed).
This PR is triggered by an error in our test suite.
```
Caused by: org.junit.platform.commons.JUnitException: Failed to parse version of junit:junit: 4.13.1
```
[The same as](https://github.com/junit-team/junit5/issues/2529).
And the recommedation is to update `junit-vintage-engine`to version >= `5.7.0`.
`jackson.version` is updated mainly due to the fact that it brings [`junit:junit:4.12 (managed)`](https://mvnrepository.com/artifact/com.fasterxml.jackson/jackson-bom/2.11.2) version with it.

Might be a replacement for #323.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,10 +45,10 @@
         <auto-matter.version>0.15.3</auto-matter.version>
         <auto-value.version>1.7.4</auto-value.version>
         <logback.version>1.2.3</logback.version>
-        <jackson.version>2.11.2</jackson.version>
+        <jackson.version>2.12.3</jackson.version>
         <semantic-metrics.version>1.1.7</semantic-metrics.version>
         <guava.version>29.0-jre</guava.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -45,10 +45,10 @@
         <auto-matter.version>0.15.3</auto-matter.version>
         <auto-value.version>1.7.4</auto-value.version>
         <logback.version>1.2.3</logback.version>
-        <jackson.version>2.11.2</jackson.version>
+        <jackson.version>2.12.3</jackson.version>
```
klaraward, line 48: In the spotify internal java/bom we currently have 2.12.2, but they should be compatible I guess.

PR #337, opened 2021-04-15
Make a HEAD /foo requests emit metrics for HEAD:/foo, not GET:/foo

mikaelgoldmann:
This fixes an issue that probably originates from the fact that
apollo registers a HEAD endpoint automatically for GET endpoints.
This means a HEAD request is served by a GET handler and endpoint
metrics are incorrectly reported for endpoint = GET:/foo rather than
HEAD:/foo.

### DIFF
--- a/modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsCollectingEndpointRunnableFactoryDecorator.java
+++ b/modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsCollectingEndpointRunnableFactoryDecorator.java
@@ -42,20 +42,13 @@ class MetricsCollectingEndpointRunnableFactoryDecorator implements EndpointRunna
   @Override
   public EndpointRunnableFactory apply(EndpointRunnableFactory delegate) {
     return (request, requestContext, endpoint) -> {
-      final String endpointName;
-      // Normally request.request().method() and endpoint.info().getRequestMethod() are equal
-      // but since apollo lets GET endpoints also serve HEAD requests, they are different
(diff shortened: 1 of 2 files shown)


modules/metrics/src/main/java/com/spotify/apollo/metrics/MetricsCollectingEndpointRunnableFactoryDecorator.java
```diff
@@ -42,7 +42,20 @@
...
+      // when the request is a HEAD request. The hack below makes sure we distinguish metrics
+      // for GET and HEAD requests.
+      // We could always take the else branch but I'm guessing that the comparison
+      // is cheaper than a needless concatenation.
+      if (request.request().method().equals(endpoint.info().getRequestMethod())) {
+        endpointName = endpoint.info().getName();
+      } else {
+        // but because GET endpoints serve HEAD requests info().getName() is
+        // wrong does not distinguish HEAD and GET requests
+        endpointName = request.request().method() + ":" + endpoint.info().getUri();
```
rydenius: Looks like this could work. But it would be more elegant if proper endpoint infos were set up for the implicit HEAD endpoints. This is a bit of a hack, isn't it?

mikaelgoldmann: Definitely a hack. I'll see if I can come up with a better option.

mikaelgoldmann:
I haven't had time for this so I will close it for now.

PR #342, opened 2021-04-21
Don't support ffwd over http anymore, use semantic-metrics 1.1.8 where the lib is not included.

klaraward:
Can not find any usages of the config `ffwd.type: http` in the Spotify code base, and I believe @lmuhlha said she wants to get rid of the ffwd-http-client library - https://github.com/spotify/apollo/pull/326#discussion_r563832060

Include the old version of `ffwd-http-client` lib to make sure downstream libraries get that as a transitive dependency and can use the shaded packages.


lmuhlha (approved):

klaraward:
We still have a problem with using a new version of semantic-metrics, because it brings in the new ffwd-http-client without the shaded dependencies, that too many people depend on.

lepistone (approved):
Sounds great together with https://github.com/spotify/semantic-metrics/pull/108. Thank you!

.circleci/config.yml
```diff
@@ -19,7 +19,7 @@ jobs:
       - run:
           command: mvn -Pcoverage -Pmissinglink verify
           environment:
-            MAVEN_OPTS: -Xmx512m
+            MAVEN_OPTS: -Xmx768m
```
toweroy, line 22: Why is this increase required?

pom.xml
```diff
@@ -162,19 +162,20 @@
             </dependency>
             <dependency>
                 <groupId>com.spotify.metrics</groupId>
-                <artifactId>semantic-metrics-core</artifactId>
+                <artifactId>semantic-metrics-bom</artifactId>
                 <version>${semantic-metrics.version}</version>
```
toweroy, line 166: If we will use the `semantic-metrics-bom` now, is this parameter required?

.circleci/config.yml
klaraward: Failed because of OOM before

PR #345, opened 2021-04-22
[metrics] Use semantic-metrics 1.1.1, revert flush function

klaraward:
Reverts #326 due to problems with downstream libraries using the shades packages in ffwd-http-client 0.4.0.
Using version 1.1.1 which seems to be the last version to include ffwd-http-client 0.4.0.

Will include newer version and reenable the flushing once we have moved downstreams deps away from importing the shaded packages

### DIFF
--- a/modules/metrics/README.md
+++ b/modules/metrics/README.md
@@ -157,7 +157,6 @@ will then emit a '0' value until the first time a certain status code shows up.
 key | type | required | note
 --- | ---- | -------- | ----
 `ffwd.interval` | int | optional | interval in seconds of reporting metrics to ffwd; default 30
-`ffwd.flush` | boolean | optional | include a final flush of metrics on service shut down; default `False`
 `ffwd.host` | string | optional | host where the ffwd agent is running. default `localhost`
 `ffwd.port` | int | optional | port where the ffwd agent is running. default `19091`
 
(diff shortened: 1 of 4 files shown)


sming (commented):
LGTM but I’m not knowledgeable about apollo and only a bit about ffwd and not at all about the context of this PR I’m afraid. I’ll defer to @lmuhlha to review in earnest if that’s cool?

lmuhlha:
Just a headsup that this will also revert:
- sprawling keys in apollo, if anyone updated graphs to not include sprawling keys they will go back to the sprawling version
- stopWithFlush which was asked for by @irock and potentially used 
- support to semantic-metrics FfwdhttpReporter (I believe the rest was in v 1.1.0) @spotify/prism

klaraward:
> Just a headsup that this will also revert:
> 
> * sprawling keys in apollo, if anyone updated graphs to not include sprawling keys they will go back to the sprawling version
> * stopWithFlush which was asked for by @irock and potentially used
> * support to semantic-metrics FfwdhttpReporter (I believe the rest was in v 1.1.0) @spotify/prism

This is definitely unfortunate. The version of Apollo that is likely to be used by most people is 1.17.4, which does not include any of these things.

toweroy:
Would users of these functions be able to do a dependency pinning in their pom file to get around the new version? They won't get subsequent updates but at least it wouldn't block them from using these new features.

PR #348, opened 2021-04-28
Add support for using either a pre- or post-closer

spkrka:


caesar-ralf:
This has moved to the internal repository

PR #356, opened 2021-05-21
Upgrade guice 4.2.2 -> 5.0.1

### LINKED ISSUE #350: Upgrade to Guice 5
Guice 5.0.1 is out! https://github.com/google/guice/wiki/Guice501

We're hoping to get rid of the spurious warnings: https://github.com/spotify/apollo/issues/246

caesar-ralf:
Upgrades Guice from 4.2.2 to 5.0.1.

# Why?

User @backuitist asked at issue https://github.com/spotify/apollo/issues/350 to do this change to remove a warning. We realized after investigation that guice 4.2.2 is not compatible with Java 16+, which will require us to do this upgrade to 5.0.1 anyway.

# Breaking changes

From documentation from their [release notes](https://github.com/google/guice/wiki/Guice501) and local tests, there's no breaking changes that we could perceive in Apollo code and its modules besides changes in dependencies.

This upgrade anyway poses a breaking change risk for users using Guice, which will be obligated to also upgrade their versions to match. Our rollout strategy has not been yet determined for this change.

## UPDATE

After testing with different projects, there's no change in the API of how Apollo is exposed, neither we could find a significant change in how Guice is used. This means that there's no breaking change perceived in both API and behaviour. With that in mind, I think we can go ahead and release this change to our users as a minor version upgrade

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,8 +75,10 @@
             </dependency>
             <dependency>
                 <groupId>com.google.inject</groupId>
-                <artifactId>guice</artifactId>
+                <artifactId>guice-bom</artifactId>
                 <version>5.0.1</version>
+                <scope>import</scope>
+                <type>pom</type>
             </dependency>
             <dependency>
                 <groupId>net.sf.jopt-simple</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -328,7 +328,7 @@
                 <plugin>
                     <groupId>com.spotify</groupId>
                     <artifactId>missinglink-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -76,12 +76,7 @@
             <dependency>
                 <groupId>com.google.inject</groupId>
                 <artifactId>guice</artifactId>
```
klaraward: Use the bom?

caesar-ralf: done

klaraward (approved):
Lets make an alpha release of this

PR #357, opened 2021-05-24
Manage lifecycle of installed modules

i-maravic:
When an ApolloModule installs another ApolloModule, it should start managing lifecycles for all the keys from the installed ApolloModule.

Apollo will not manage the installed modules on the service shutdown if we don't do this.

### DIFF
--- a/apollo-core/src/test/java/com/spotify/apollo/core/ServiceImplTest.java
+++ b/apollo-core/src/test/java/com/spotify/apollo/core/ServiceImplTest.java
@@ -447,6 +447,27 @@ public class ServiceImplTest {
   }
 
   @Test
+  public void testChildModuleClassesAreLifecycleManaged() throws Exception {
+    AtomicBoolean created = new AtomicBoolean(false);
+    AtomicBoolean closed = new AtomicBoolean(false);
+    AtomicBoolean childCreated = new AtomicBoolean(false);
+    AtomicBoolean childClosed = new AtomicBoolean(false);
+    ModuleWithLifecycledKeys lifecycleModule =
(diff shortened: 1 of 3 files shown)


i-maravic:
@lepistone

apollo-core/src/test/java/com/spotify/apollo/core/ServiceImplTest.java
```diff
@@ -446,6 +446,27 @@ public void testModuleClassesAreLifecycleManaged() throws Exception {
...
+    Service service = ServiceImpl.builder("test")
+        .withModule(lifecycleModule)
+        .build();
+
+    try (Service.Instance instance = service.start()) {
+      instance.getSignaller().signalShutdown();
+      instance.waitForShutdown();
+    }
+
+    assertThat(childCreated.get(), is(true));
```
caesar-ralf, line 466: nitty-pick: What do you think about repeating this at line 462? Something like this could be possible:

```
try (Service.Instance instance = service.start()) {
    assertThat(childCreated.get(), is(true));
    assertThat(childClosed.get(), is(false));
    instance.getSignaller().signalShutdown();  
    instance.waitForShutdown();    
}
```

From what I am looking in the code I think it should be fine as they are set in the moment you're creating the instance of the class. What do you think?

klaraward:
Reviewed and approved by weaver squad.
&lt;jinx&gt;Since the keys are stored in a ImmutableSet and then a LinkedHashSet we are positive the ordering of the lifecycle managed keys is preserved and this PR should not lead to any big surprises in the shutdown sequence of an Apollo service.&lt;/jinx&gt;

klaraward (approved):

PR #359, opened 2021-07-02
Trying com.spotify:dns 3.2.2 with dnsjava 3.0.2

klaraward:
See internal PR https://ghe.spotify.net/java/bom/pull/538


caesar-ralf:
We merged PR that removes the need for Spotify DNS as it's not being used

Issue #362, opened 2021-08-18
Modules with decorators are restricted by order they are declared


caesar-ralf:
# The problem

User wanted to add metrics to their http outgoing calls. To do that they decided to add `HttpMetricModule` in their service like so:

```
   final Service service =
          Services.usingName(SERVICE_NAME)
              .withModule(StandaloneModule.create(Main::configure))
              .withEnvVarPrefix("SPOTIFY")
              **.withModule(HttpMetricModule.create())**
              .withModule(HttpClientModule.create())
              .build();
```

Then when they did their calls they noticed that there was no metrics being sent.

# What happened?

When we investigate the error we figured out that `HttpMetricModules` create a decorator for the client class, but the client class was not being decorated, even with the module declared.

While for most of the cases ordering of the modules doesn't matter as Guice will lazily resolve bindings, for Decorators it does matter. Unless the decorator is explicitly set after what is decorating, the decoration won't work as expected.

# What we expect?

We can try to solve this issue directly in Apollo. As users currently are declaring all their modules in a single place, we can try some different approaches:

* Force `priority` of decorators to be less than what they are decorating
* Create hint for decorator modules with something like `@InstallAfter` , `@DependsOn`, or something similar
* others?

If not solvable, we need to make it clear somehow for users that this problem might ocurr somehow, either by documentation, or emitting warnings.

caesar-ralf:
This has moved to the internal repository

PR #363, opened 2021-08-18
Adds more documentation about http modules

caesar-ralf:
To improve the situation due to issue #362


modules/okhttp-client/src/main/java/com/spotify/apollo/http/client/HttpMetricModule.java
```diff
@@ -28,6 +27,35 @@
...
 
+/**
+ * Module that adds metrics to any http call using the Apollo client.
+ *
+ * <p>The declaration of this module must be done <b>AFTER</b> the {@link HttpClientModule} like so:
+ *
+ * <blockquote>
+ *
+ * See <a href="https://github.com/spotify/apollo/issues/362">issue#362</a> for more information *
+ * of why the order matters.
```
klaraward:
```suggestion
 * on why the order matters.
```

klaraward (approved):

PR #364, opened 2021-08-19
Bom should not have a parent

klaraward:
Solving #365 

One particular dependency that users of apollo-bom might have expected to get the version for is the `ffwd-http-client`.
Internally in Spotify this will be solved by defining that version in the internal Spotify bom.


lepistone (approved):
Thanks!

toweroy (approved):

PR #366, opened 2021-08-24
Add possibility to use Spring Boot starters

herder:
This change means that you can utilize the Spring Boot starters (spring-cloud-gcp-*-starter, for example), to access the beans that they create. So you could use the Spring Boot pubsub support for example, by getting a handle to a PubsubTemplate via environment.resolve(PubsubTemplate.class)

One big limitation right now is that you can't define anything in your own app, so creating your own `@Configuration` classes for wiring your own beans in your app doesn't work, and creating Spring Data repositories won't work. The difficult part is trying to achieve this without the client code needing any changes - I think it means that we would need to change in the `StandaloneService` code so that it kicks off the autoconfiguration for the Service class that's provided to it. But I haven't tried that yet.

### DIFF
--- a/apollo-core/src/main/java/com/spotify/apollo/core/ApolloInitializer.java
+++ b/apollo-core/src/main/java/com/spotify/apollo/core/ApolloInitializer.java
@@ -1,0 +1,72 @@
+/*
+ * -\-\-
+ * Spotify Apollo Service Core (aka Leto)
+ * --
+ * Copyright (C) 2013 - 2021 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+package com.spotify.apollo.core;
+
+import com.google.inject.Module;
+import com.spotify.apollo.module.ApolloModule;
+import org.springframework.context.ApplicationContextInitializer;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.core.env.PropertySource;
+
+public class ApolloInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+
+
+  private final String serviceName;
+  private final Iterable<Module> guiceModules;
+
+  public ApolloInitializer(String serviceName, Iterable<Module> guiceModules) {
+    this.serviceName = serviceName;
+    this.guiceModules = guiceModules;
+  }
+
+  @Override
+  public void initialize(
+      ConfigurableApplicationContext configurableApplicationContext) {
+    configurableApplicationContext.getEnvironment().getPropertySources()
+        .addFirst(new ApolloPropertySource(serviceName));
+    configurableApplicationContext.getEnvironment().getPropertySources()
+        .addFirst(new PropertySource<String>("serviceName") {
+          @Override
+          public Object getProperty(String name) {
+            return "spring.application.name".equals(name) ? serviceName : null;
+          }
+        });
+
+    for (Module allModule : guiceModules) {
+      configurableApplicationContext.getBeanFactory().registerResolvableDependency(allModule.getClass(), allModule);
+      String id = allModule instanceof ApolloModule ? ((ApolloModule) allModule).getId() : allModule.toString();
+        configurableApplicationContext.getBeanFactory().registerSingleton(findUniqueId(id, configurableApplicationContext), allModule);
+    }
+  }
+
+  private String findUniqueId(String id,
+                              ConfigurableApplicationContext configurableApplicationContext) {
+    if (!configurableApplicationContext.containsBean(id)) {
+      return id;
+    } else {
+      for(int i = 1;; i++) {
+        String newId  = id + "-" + i;
+        if (!configurableApplicationContext.containsBean(newId)) {
+          return newId;
+        }
+      }
+    }
+  }
+}
--- a/apollo-core/src/main/java/com/spotify/apollo/core/ApolloInitializer.java
+++ b/apollo-core/src/main/java/com/spotify/apollo/core/ApolloInitializer.java
@@ -21,12 +21,16 @@ package com.spotify.apollo.core;
 
 import com.google.inject.Module;
 import com.spotify.apollo.module.ApolloModule;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.context.ApplicationContextInitializer;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.core.env.PropertySource;
 
-public class ApolloInitializer implements ApplicationContextInitializer<ConfigurableApplicationContext> {
+public class ApolloInitializer
+    implements ApplicationContextInitializer<ConfigurableApplicationContext> {
 
+  private static final Logger LOGGER = LoggerFactory.getLogger(ApolloInitializer.class);
 
   private final String serviceName;
   private final Iterable<Module> guiceModules;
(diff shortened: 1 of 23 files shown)


pom.xml
```diff
@@ -51,6 +51,7 @@
         <junit.testkit.version>1.7.1</junit.testkit.version>
         <slf4j.version>1.7.30</slf4j.version>
         <hamcrest.version>2.2</hamcrest.version>
+        <spring.version>2.5.3</spring.version>
```
Mandalorian007: I think I would opt for labeling this as the spring-boot version over the Spring version

apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
```diff
@@ -207,12 +214,27 @@ InstanceImpl initInstance(
...
+
+    final SpringApplicationBuilder springApplicationBuilder = new SpringApplicationBuilder()
+        .properties(
+           //Prefer Spring bean if present in both Guice and application context
+            "spring.guice.dedup=true",
+           //Do not create default instances if there is no instance in the context
+           "spring.guice.autowireJIT=false",
+          //Allow overriding of bean definitions from different modules
+           "spring.main.allow-bean-definition-overriding=true")
+        .sources(GuiceSpringBridge.class)
```
Mandalorian007: Since you are adding this class as a source I am fairly certain you can leverage the @ComponentScan annotation on this class to enable picking up @Configuration, @Bean, and all variations of @Component

Mandalorian007 (commented):
Looks like a solid integration start I will have to pull it down and test it, but added a few small comments --mcf

pom.xml
herder: Done in fc3d1bda1b0a6b582cf16e46b01df1fe22eda4e9

apollo-core/src/main/java/com/spotify/apollo/core/ApolloInitializer.java
```diff
@@ -0,0 +1,92 @@
...
+          .getBeanFactory()
+          .registerResolvableDependency(eachModule.getClass(), eachModule);
+      String id =
+          eachModule instanceof ApolloModule
+              ? ((ApolloModule) eachModule).getId()
+              : eachModule.toString();
+      configurableApplicationContext
+          .getBeanFactory()
+          .registerSingleton(findUniqueId(id, configurableApplicationContext), eachModule);
+      LOGGER.info(
```
Mandalorian007: This could get noisy. Would it make more sense to move this info to debug and maybe add a message that reports how many items were added to the context on info?

apollo-core/src/main/java/com/spotify/apollo/core/ServiceImpl.java
helenaAbergOstlund: Thanks for the tip! This is working now after some struggles.

helenaAbergOstlund:
> One big limitation right now is that you can't define anything in your own app, so creating your own `@Configuration` classes for wiring your own beans in your app doesn't work, and creating Spring Data repositories won't work. The difficult part is trying to achieve this without the client code needing any changes - I think it means that we would need to change in the `StandaloneService` code so that it kicks off the autoconfiguration for the Service class that's provided to it. But I haven't tried that yet.

This has now been solved.

apollo-core/src/main/java/com/spotify/apollo/core/GuiceSpringBridge.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.guice.annotation.EnableGuiceModules;
+
+/** Glue to knit together Guice, Spring and make the Spring component scan work. */
+@Configuration
+@EnableGuiceModules
+@SpringBootConfiguration
+@EnableAutoConfiguration
+@ComponentScan(basePackages = "com.spotify")
```
Mandalorian007: hmm so this will function fo any project that leverages a `com.spotify` package which places a bit of an interesting restriction on Apollo projects that would want to use Spring needing to have this package structure. Additionally, the usage of the `@EnableAutoConfiguration` at this package would treat the current package structure as the root package and would only configure the application context with things contained beneath it.

One idea that might actually make this a bit cleaner would be to have an @SpringBootApplication declared on the Main class within the Apollo Standalone (this annotation packages @Configuration, @EnableAutoConfiguration and @ComponentScan). Then since we need to register the main class as a source maybe we could leverage java reflections to search for any class annotated with @SpringBootApplication and register it.

Mandalorian007 (commented):
Added some additional thoughts on Autowiring

apollo-core/src/main/java/com/spotify/apollo/core/ApolloInitializer.java
helenaAbergOstlund: Figured it was ok since Apollo code also logs all loaded modules on INFO level.

apollo-core/src/main/java/com/spotify/apollo/core/GuiceSpringBridge.java
helenaAbergOstlund: Unfortunately I cannot get it to work without the `basePackages = "com.spotify"`. 

Your idea might work better but not something I will have time to try now as the hackweek is over.

apollo-core/src/main/java/com/spotify/apollo/core/ApolloInitializer.java
Mandalorian007: All good then this will work it was just a thought it could add a bunch of logs

helenaAbergOstlund:
Fixed conflicts with 1.x branch and the jdk8 build.

apollo-core/src/main/java/com/spotify/apollo/core/GuiceSpringBridge.java
herder: I took a stab at fixing this a more generic way in 81cccfb
This makes the Spring bridge look for a `@SpringBootApplication`-annotated class to add as a source when initializing. 
See commit message for some questions, @helenaAbergOstlund @Mandalorian007

helenaAbergOstlund: I like it!

> The main drawback is that the `main` class needs to be annotated with `@SpringBootApplication`, but on the other hand this makes the application a more by-the-book Spring Boot app.
> On the other hand this could be a feature: if we find an app annotated with `@SpringBootApplication` we auto-activate the Spring support, only skipping that if `apollo.spring.enabled` is explicitly set to `false`

Opting for the latter - assuming an app annotated with '@SpringBootApplication' will want to enable spring support - makes a lot of sense. I guess we need to change the code in that case so we always check the stack trace and scan for the annotation, and not only when the flag is enabled.

herder: > Opting for the latter - assuming an app annotated with '@SpringBootApplication' will want to enable spring support - makes a lot of sense. I guess we need to change the code in that case so we always check the stack trace and scan for the annotation, and not only when the flag is enabled.

I agree: it makes more sense to try to follow the normal way of activating Spring as far as possible. One question is how to make this work with Fleet Management, though - it would require code changes besides just version bumping to get Spring support in all services. That might not be a real problem, though - I assume it's not too hard to create a fleet change that inspects the main class and adds the annotation if required. But this might be a problem for the future either way - we might cross that bridge when we need it :)

herder:
I changed the activation strategy to how I interpreted our discussion [here](https://github.com/spotify/apollo/pull/366#discussion_r702826082) so that Spring support is activated if one of these conditions is fulfilled:

- The main class is annotated with `@SpringBootApplication` and `apollo.spring.enabled` is not present or not set to `false`
- The `@SpringBootApplication` class is present on the classpath and `apollo.spring.enabled` is not set to `false`

https://github.com/spotify/apollo/pull/366/commits/47b51dedc77d76e3176e9600b7d2cff537cce6bb

Also, all the Spring dependencies are now optional except for `spring-guice`. To me it fits better with the 'just works' ideal that you only need to add the regular Spring dependencies if you want to make Apollo into a Spring app - you shouldn't have to care about how it is actually accomplished. And the classes are only loaded if Spring support is activated.

caesar-ralf:
This has moved to the internal repository

PR #367, opened 2021-08-26
Use properties for distributionManagement

klaraward:


pom.xml
```diff
@@ -669,16 +674,17 @@
                 <maven.compiler.release>8</maven.compiler.release>
             </properties>
         </profile>
+
```
klaraward:
```suggestion
```

toweroy (approved):

